PR #6, opened 2019-03-07
WPS1 support

fmigneault:


weaver/tests/functional/application-packages/WPS1/DeployProcess_CatalogSearch.json
```diff
@@ -38,8 +44,8 @@
     },
     "executionUnit": [
         {
-            "href": "catalog_search.cwl"
+            "href": "weaver/tests/functional/application-packages/WPS1/catalog_search.cwl"
         }
     ],
-    "deploymentProfileName": "http://www.opengis.net/profiles/eoc/workflow"
+    "deploymentProfileName": "http://www.opengis.net/profiles/eoc/wpsApplication"
```
dbyrns: I know that the only options that we have right are "workflow" and "dockerizedApplication", so wpsApplication seems to make sense... But are we actually using that value?

weaver/wps_restapi/processes/processes.py
```diff
@@ -518,7 +520,9 @@ def add_local_process(request):
             raise HTTPUnprocessableEntity("Invalid parameter 'processDescription.process.owsContext.offering.content'.")
         package = None
         reference = content.get('href')
-    elif deployment_profile_name.endswith('workflow') or deployment_profile_name.endswith('application'):
+    elif deployment_profile_name:
+        if not any(deployment_profile_name.endswith(typ) for typ in [PROCESS_APPLICATION, PROCESS_WORKFLOW]):
```
dbyrns: I guess that I found the answer myself... So Dockerized and WpsApplication both end with application (but endswith isn't case sensitive?)

weaver/tests/functional/application-packages/WPS1/DeployProcess_CatalogSearch.json
fmigneault: yes, see your answer ;)

weaver/wps_restapi/processes/processes.py
fmigneault: there's a lower a few lines above

Issue #5, opened 2019-03-06
simplify the whole projet


fmigneault:
remove supervisor nginx mongodb // only run pyramid app

will save a LOT of space for image size
will simplify by a LOT the image/config setup

## todo

- remove buildout and related mongodb, supervisor, etc.. parts
- use directly gunicorn instance in docker
- link to external mongodb
- remove installs in following make according to removed dependencies
```
Makefile:181
"$(CONDA_HOME)/bin/conda" install -y -n "$(CONDA_ENV)" "setuptools=$(SETUPTOOLS_VERSION)" supervisor nginx mongodb
```
- figure out how to handle celery... external one would duplicate code a lot...
- figure out how to serve pywps endpoint, pywps-data-output dir endpoint and weaver app all under the same pyramid app (maybe using `[composite:main]` in `weaver.ini` ?)
- update compose of other repos to update `weaver.ini` file directly instead of `custom.cfg` since it won't be used anymore + add link / other images required

davidcaron:
Couldn't agree more. That will help development and deployment a lot.

davidcaron:
FYI bird-house/twitcher doesn't use buildout anymore.

fmigneault:
fixed by #59

PR #23, opened 2019-03-19
Esgf cwt process

davidcaron:
- [X] Executing the aggregate process using boreas data
- [X] Executing the subset process using boreas data
- [X] Add a unit test for subset (aggregate wouldn't be necessary, as subset requires more parameters)

(https://aims2.llnl.gov/wps/ is down)
- [ ] Executing the aggregate process using esgf data
- [ ] Executing the subset process using esgf data

The unchecked tasks will be in a different PR.


davidcaron:
@fmigneault This branch would be ready to merge. I'll wait for your OK, because there are some things you may want to take a look at that should have been in another PR:

- in `weaver/processes/wps1_process.py` I changed some logic when converting from cwl inputs to pywps
- in `weaver/wps_restapi/processes/processes.py` I added a `reset_connection` argument, to be called from the forked celery process. I was getting a connection error from pymongo when I was accessing the connection for the first time in the celery worker.

davidcaron:
@fmigneault I would be ready to merge this, let me know if you have any questions.

davidcaron:
the process `file2string_array.py` to convert a file input into a string array works as expected. 

I added a simple test case that runs a workflow: File -> string array -> subset colibri

fmigneault:
@davidcaron @dbyrns 
don't know what is causing it, but many tests fail.
they are all passing on master

setup.cfg
```diff
@@ -2,7 +2,7 @@
 addopts =
         --strict
         --tb=native
-        weaver/
+;        weaver/
```
fmigneault: shouldn't be removed, `make test` calls depend on it

tests/functional/application-packages/DeployProcess_AggregateESGF.json
```diff
@@ -0,0 +1,44 @@
...
+                            "glob": "output_netcdf.nc"
+                        },
+                        "type": "File"
+                    }
+                }
+            }
+        }
+    ],
+    "deploymentProfileName": "http://www.opengis.net/profiles/eoc/ESGFWpsApplication"
+}
```
fmigneault: add newline

tests/functional/application-packages/DeployProcess_SubsetESGF.json
```diff
@@ -0,0 +1,75 @@
...
+                            "glob": "output_netcdf.nc"
+                        },
+                        "type": "File"
+                    }
+                }
+            }
+        }
+    ],
+    "deploymentProfileName": "http://www.opengis.net/profiles/eoc/ESGFWpsApplication"
+}
```
fmigneault: idem

tests/functional/application-packages/DeployProcess_SubsetNASAESGF.json
```diff
@@ -0,0 +1,72 @@
...
+                            "glob": "output_netcdf.nc"
+                        },
+                        "type": "File"
+                    }
+                }
+            }
+        }
+    ],
+    "deploymentProfileName": "http://www.opengis.net/profiles/eoc/ESGFWpsApplication"
+}
```
fmigneault: idem

tests/functional/application-packages/Execute_AggregateESGF.json
```diff
@@ -0,0 +1,28 @@
...
+			"data": "{{ api_key }}"
+		}
+  ],
+  "outputs": [
+    {
+      "id": "output",
+      "transmissionMode": "reference"
+    }
+  ]
+}
```
fmigneault: idem

tests/functional/application-packages/Execute_SubsetESGF.json
```diff
@@ -0,0 +1,60 @@
...
+			"data": "values"
+		}
+  ],
+  "outputs": [
+    {
+      "id": "output",
+      "transmissionMode": "reference"
+    }
+  ]
+}
```
fmigneault: idem

tests/functional/application-packages/Execute_SubsetNASAESGF.json
```diff
@@ -0,0 +1,56 @@
...
+			"data": "values"
+		}
+  ],
+  "outputs": [
+    {
+      "id": "output",
+      "transmissionMode": "reference"
+    }
+  ]
+}
```
fmigneault: idem

tests/functional/application-packages/Execute_WorkflowFile_To_SubsetCRIM.json
```diff
@@ -0,0 +1,32 @@
...
+      "data": "-60"
+    }
+  ],
+  "outputs": [
+    {
+      "id": "output",
+      "transmissionMode": "reference"
+    }
+  ]
+}
```
fmigneault: idem

tests/functional/application-packages/Execute_WorkflowSubsetLLNL_SubsetCRIM.json
```diff
@@ -3,39 +3,48 @@
...
-      "id": "llnl_lon0",
-      "data": "-128"
+      "id": "variable",
+      "data": "tasmin"
     },
     {
-      "id": "llnl_lon1",
-      "data": "-48"
+      "id": "esgf_api_key",
+      "data": "G2LhbLn4clvTW9SHM10D0ZGPFPkEeCmm7zJU0LPJ7mTaJzrgkJ3tazcsxdCp3nl5"
```
fmigneault: Good chance this shouldn't be here as a literal string
We could use a Template string, and replace with an env var.
For test on travis, the value should be pushed as an encrypted value, which is specified in .travis.yml

tests/functional/application-packages/Execute_WorkflowSubsetNASAESGF_SubsetCRIM.json
```diff
@@ -0,0 +1,52 @@
...
+      "data": "-65"
+    }
+  ],
+  "outputs": [
+    {
+      "id": "output",
+      "transmissionMode": "reference"
+    }
+  ]
+}
```
fmigneault: newline

tests/functional/test_ems_end2end.py
```diff
@@ -219,23 +219,35 @@ def setup_test_processes(cls):
...
+        cls.PROCESS_SUBSET_ESGF = "SubsetESGF"
+        cls.PROCESS_SUBSET_NASAESGF = "SubsetNASAESGF"
         cls.PROCESS_WORKFLOW_ID = "Workflow"
         cls.PROCESS_WORKFLOW_SC_ID = "WorkflowSimpleChain"
         cls.PROCESS_WORKFLOW_S2P_ID = "WorkflowS2ProbaV"
         cls.PROCESS_WORKFLOW_CUSTOM_ID = "CustomWorkflow"
         cls.PROCESS_WORKFLOW_FLOOD_DETECTION_ID = "WorkflowFloodDetection"
         cls.PROCESS_WORKFLOW_SUBSET_ICE_DAYS = "WorkflowSubsetIceDays"
+        cls.PROCESS_WORKFLOW_SubsetLLNL_SubsetCRIM = "WorkflowSubsetLLNL_SubsetCRIM"
+        cls.PROCESS_WORKFLOW_SubsetNASAESGF_SubsetCRIM = "WorkflowSubsetNASAESGF_SubsetCRIM"
```
fmigneault: make all caps to remain consistent in var naming, use underscore to split acronyms as necessary

weaver/database/mongodb.py
```diff
@@ -91,10 +91,12 @@ def run_migration(self):
         pass
 
 
-def get_mongodb_connection(container):
-    # type: (AnySettingsContainer) -> Database
+def get_mongodb_connection(container, reset_connection=False):
+    # type: (AnySettingsContainer, bool) -> Database
```
fmigneault: Optional[bool]

weaver/database/mongodb.py
```diff
@@ -107,10 +109,10 @@ def get_mongodb_connection(container):
     return MongoDB
 
 
-def get_mongodb_engine(container):
-    # type: (AnySettingsContainer) -> Database
+def get_mongodb_engine(container, reset_connection=False):
+    # type: (AnySettingsContainer, bool) -> Database
```
fmigneault: Optional[bool]

weaver/processes/builtin/file2string_array.py
```diff
@@ -0,0 +1,43 @@
...
+CUR_DIR = os.path.abspath(os.path.dirname(__file__))
+
+PACKAGE_NAME = os.path.split(os.path.splitext(__file__)[0])[-1]
+
+# setup logger since it is not run from the main 'weaver' app
+LOGGER = logging.getLogger(__name__)
+LOGGER.addHandler(logging.StreamHandler(sys.stdout))
+LOGGER.setLevel(logging.INFO)
+
+output_cwl_json = "cwl.output.json"
```
fmigneault: move inside function

weaver/processes/esgf_process.py
```diff
@@ -0,0 +1,254 @@
...
+from weaver.processes.wps1_process import Wps1Process
+
+if TYPE_CHECKING:
+    from weaver.typedefs import JsonBody
+    from typing import AnyStr, Dict, List, Tuple
+
+LOGGER = logging.getLogger(__name__)
+
+
+class Percent:
```
fmigneault: Percent(object)

weaver/processes/esgf_process.py
```diff
@@ -0,0 +1,254 @@
...
+    PREPARING = 2
+    SENDING = 3
+    COMPUTE_DONE = 98
+    FINISHED = 100
+
+
+class InputNames:
```
fmigneault: InputNames(object)

weaver/processes/esgf_process.py
```diff
@@ -0,0 +1,254 @@
...
+from typing import AnyStr, TYPE_CHECKING, Optional
+import logging
+import requests
+import cwt
+
+from weaver.status import STATUS_RUNNING, STATUS_SUCCEEDED, STATUS_FAILED
+from weaver.processes.wps1_process import Wps1Process
+
+if TYPE_CHECKING:
+    from weaver.typedefs import JsonBody
```
fmigneault: JsonBody => JSON
and adjust references to it

weaver/processes/esgf_process.py
```diff
@@ -0,0 +1,254 @@
...
+    time = "time"
+    lat = "lat"
+    lon = "lon"
+
+
+class ESGFProcess(Wps1Process):
+    required_inputs = ("variable", )
+
+    def execute(self, workflow_inputs, output_dir, expected_outputs):
+        # type: (JsonBody, AnyStr, Dict[AnyStr, AnyStr]) -> None
```
fmigneault: can remove, let inheritance from base class

weaver/processes/esgf_process.py
```diff
@@ -0,0 +1,254 @@
...
+        files = self._get_files_urls(workflow_inputs)
+        varname = self._get_variable(workflow_inputs)
+
+        LOGGER.debug("Creating esgf-compute-api inputs")
+
+        inputs = [cwt.Variable(url, varname) for url in files]
+
+        return inputs
+
+    def _get_domain(self, workflow_inputs):
```
fmigneault: `@staticmethod`

weaver/processes/esgf_process.py
```diff
@@ -0,0 +1,254 @@
...
+            InputNames.lat,
+            InputNames.lon,
+        ]
+
+        grouped_inputs = defaultdict(dict)
+
+        for dim_name in dimensions_names:
+            for param, v in workflow_inputs.items():
+                if param.startswith(dim_name + "_"):
+                    param_splitted = param.split("_", 1)[1]
```
fmigneault: `param_split` (spell check)

weaver/processes/esgf_process.py
```diff
@@ -0,0 +1,254 @@
...
+                dimensions
+            )
+            return domain
+
+    def _check_required_inputs(self, workflow_inputs):
+        for required_input in self.required_inputs:
+            if required_input not in workflow_inputs:
+                raise ValueError("Missing required input: {}".format(required_input))
+
+    def _get_files_urls(self, workflow_inputs):
```
fmigneault: `@staticmethod`

weaver/processes/esgf_process.py
```diff
@@ -0,0 +1,254 @@
...
+        for cwl_file in files:
+            if not cwl_file["class"] == "File":
+                raise ValueError("'{}' inputs must have a class named 'File'".format(InputNames.files))
+            location = cwl_file["location"]
+            if not location.startswith("http"):
+                raise ValueError("ESGF processes only support urls for files inputs.")
+            urls.append(location)
+        return urls
+
+    def _get_variable(self, workflow_inputs):
```
fmigneault: `@staticmethod`

weaver/processes/esgf_process.py
```diff
@@ -0,0 +1,254 @@
...
+
+        return process
+
+    def _wait(self, esgf_process, sleep_time=2):
+        # type: (cwt.Process, float) -> bool
+        """Wait for an ESGF process to finish, while reporting its status"""
+        status_history = set()
+
+        status_percent = [0]  # python 2 can't mutate nonlocal
+        last_percent_regex = re.compile(r".+ (\d{1,3})$")
```
fmigneault: move to top of file as `LAST_PERCENT_REGEX`
no point to compile on each call

weaver/processes/esgf_process.py
```diff
@@ -0,0 +1,254 @@
+import re
+import time
+from collections import defaultdict
+from os.path import join
+
+from typing import AnyStr, TYPE_CHECKING, Optional
+import logging
+import requests
+import cwt
```
fmigneault: requirement? getting unresolved ref error

weaver/processes/esgf_process.py
```diff
@@ -0,0 +1,254 @@
+import re
+import time
+from collections import defaultdict
+from os.path import join
+
+from typing import AnyStr, TYPE_CHECKING, Optional
```
fmigneault: only TYPE_CHECKING

weaver/processes/esgf_process.py
```diff
@@ -0,0 +1,254 @@
+import re
```
fmigneault: keep same structure order as other files

```
from weaver ...
from other...
import ...
```

weaver/processes/esgf_process.py
```diff
@@ -0,0 +1,254 @@
...
+
+    def _write_outputs(self, url, output_dir, expected_outputs):
+        """Write the output netcdf url to a local drive"""
+        message = "Downloading outputs."
+        LOGGER.debug(message)
+        self.update_status(message, Percent.COMPUTE_DONE, STATUS_RUNNING)
+
+        nc_outputs = [v for v in expected_outputs.values() if v.lower().endswith(".nc")]
+        if len(nc_outputs) > 1:
+            raise NotImplemented("Multiple outputs are not implemented")
```
fmigneault: should be `NotImplementedError`

weaver/wps_restapi/processes/processes.py
```diff
@@ -22,7 +23,8 @@
     STATUS_SUCCEEDED,
 )
 from weaver.store.base import StoreServices, StoreProcesses, StoreJobs
-from weaver.utils import get_any_id, get_any_value, get_settings, raise_on_xml_exception, wait_secs, get_cookie_headers
+from weaver.utils import get_any_id, get_any_value, get_settings, raise_on_xml_exception, wait_secs, get_cookie_headers, \
+    get_registry
```
fmigneault: too long change to (newlines as required)
```
from weaver.utils import (
    get_any_id, get_any_value, get_settings, raise_on_xml_exception, 
    wait_secs, get_cookie_headers, get_registry
)
```

weaver/processes/wps_package.py
```diff
@@ -1634,23 +1634,37 @@ def _update_status_dispatch(_provider, _message, _progress, _status):
...
-                             "only one permitted amongst {}.".format(list(app_hints), list(CWL_REQUIREMENT_APP_TYPES)))
-        requirement = app_hints[0] if app_hints else {"class": ""}
-        if requirement["class"].endswith(CWL_REQUIREMENT_APP_WPS1):
-            req_params = ["provider", "process"]
-            if not all(r in requirement for r in ["provider", "process"]):
-                raise ValueError("Missing requirement [{}] details amongst {}".format(requirement["class"], req_params))
-            provider = requirement["provider"]
-            # The process id of the provider isn't required to be the same as the one use in the EMS
-            process = requirement["process"]
+                             "only one permitted amongst {}.".format(list(app_hints), PACKAGE_REQUIREMENTS_APP_TYPES))
```
fmigneault: PACKAGE_REQUIREMENTS_APP_TYPES => CWL_REQUIREMENT_APP_TYPES

weaver/wps_restapi/processes/processes.py
```diff
@@ -70,7 +72,10 @@ def execute_process(self, job_id, url, headers=None, notification_email=None):
     load_pywps_cfg(settings)
 
     ssl_verify = asbool(settings.get("weaver.ssl_verify", True))
-    store = get_db(app).get_store(StoreJobs)
+
+    # reset the connection because we are in a forked celery process
+    db = MongoDatabase(get_registry(app), reset_connection=True)
+    store = db.get_store(StoreJobs)
```
fmigneault: revert to `get_db`

weaver/wps_restapi/processes/processes.py
```diff
@@ -1,5 +1,6 @@
 from weaver.config import get_weaver_configuration, WEAVER_CONFIGURATION_EMS
 from weaver.database import get_db
+from weaver.database.mongodb import MongoDatabase
```
fmigneault: remove

weaver/wps.py
```diff
@@ -185,4 +185,7 @@ def includeme(config):
         wps_path = get_wps_path(settings)
         config.add_route("wps", wps_path)
         config.add_view(pywps_view, route_name="wps")
+
+        config.add_static_view(get_wps_output_path(config), get_wps_output_dir(config))
```
fmigneault: this shouldn't be required
will cause problems with nginx that is currently doing it

tests/functional/application-packages/Execute_WorkflowFile_To_SubsetCRIM.json
```diff
@@ -0,0 +1,32 @@
+{
+  "mode": "async",
+  "response": "document",
+  "inputs": [
+    {
+      "id": "file",
+      "href": "https://github.com/Ouranosinc/xclim/raw/master/tests/testdata/HadGEM2-CC_360day/tasmax_day_HadGEM2-CC_rcp85_r1i1p1_na10kgrid_qm-moving-50bins-detrend_2095.nc"
```
fmigneault: use a commit id so that we keep the same reference, so avoid suddenly changing test result if this is modified

setup.cfg
davidcaron: yes, sorry, I do that to run a single test from pytest. Do you have another method to target a single test? when this line is there, The whole test suite is ran.

weaver/wps_restapi/processes/processes.py
davidcaron: maybe we should have a linter running on the ci...

weaver/processes/wps_package.py
davidcaron: thanks

weaver/wps.py
davidcaron: Wouldn' t nginx override this? The way I see it it doesn't cause any harm because nginx catches the route before weaver does. And adding this makes running a local weaver without the need to have nginx running.

weaver/wps_restapi/processes/processes.py
davidcaron: I think in this context we need to reset the mongodb connection because the celery worker is forked, and the connection can't be shared. I was getting connection errors when I didn't do it. Do you have any idea why I would get connection errors?

weaver/wps_restapi/processes/processes.py
davidcaron: see previous comment

weaver/wps_restapi/processes/processes.py
davidcaron: see: http://api.mongodb.com/python/current/faq.html#pymongo-fork-safe

davidcaron:
aims2 is up this morning. I will test the workflows requiring aims2.

weaver/wps.py
dbyrns: Good point David.

davidcaron: Also, isn't this the idea for #5?

fmigneault: If it works in both cases, great.
Actually, we could remove it altogether from nginx, because `weaver.wps_output_dir` and `weaver.wps_output_path` are hardcode in it.
This variant is better if workflows are functional with it.

weaver/wps_restapi/processes/processes.py
fmigneault: Maybe cause by the various connections between celery/weaver when entering this module?
I have no issue with the `reset_connection=True`, but I would prefer to have it as an input of `get_db`, and this one could have a `*args, **kwargs` in it signature, that it passed down to `MongoDatabase`.
I want to have all calls to db connections regrouped in this function. Otherwise we end up having patches like this left and right.

weaver/wps.py
davidcaron: nginx would serve the static content 'faster' but I don't know by how much.

setup.cfg
fmigneault: In pycharm, there's a green arrow next to a test, just press it and it creates a test config.
Usually, I just need to change the working dir in it to point to weaver root and all works.

weaver/wps.py
fmigneault: Still worth it to move toward #5. So let's go with your solution and directly remove the config in nginx while we are at it.

dbyrns: I do not agree to remove the file serving by nginx. We are talking of huge netcdf file and we should avoid overhead here...

fmigneault: But it could be a problem, since one could set `weaver.wps_output_path = "/my-new-path` and wonder for a long time why behavior is not as expected.

davidcaron: I didn't think about large files... wouldn't nginx be more appropriate for large files? 

Also, wouldn't setting `weaver.wps_output_path = "/my-new-path"` break the correct saving of outputs anyway?

I think we can have both... keep the static route in weaver, remove nginx in the config, but when we deploy, configure another instance of nginx to serve static content on a specific route.

weaver/wps_restapi/processes/processes.py
davidcaron: see 8ac83b0

fmigneault: need the Optional[] typing

weaver/wps.py
fmigneault: The idea with `weaver.wps_output_path` is to specify which weaver url to call to obtain the process results. I couldn't find a way to cleanly match nginx's, pywps's and weaver's configs together under this same setting. So currently, it just "works" because the values are all the same...
I managed to somehow patch the pywps/weaver configs with this setting, but not nginx. I was counting on #5 to just drop it.

dbyrns: There's not only large file, there is simultaneously access... Nginx will serve multiples files at the same time without issues while we will rapidly run out of python workers all strugling to serve the files.
For the config issue why not using an environment variable both used in the nginx and weaver config files with the template setup we have?

fmigneault: An external nginx in the compose to serve the files is an interesting idea.
We would need to make sure that `add_static_view` works standalone so that anyone can still run weaver by itself.

weaver/wps_restapi/processes/processes.py
davidcaron: doesn't Optional mean it can be None?

davidcaron: see: https://docs.python.org/3/library/typing.html#typing.Optional

fmigneault: oh, the more you know... will need to revert some typing in many places!

weaver/wps.py
dbyrns: The external nginx is already serving the files using the volume wps_outputs. 
so I think we got a consensus that #5 is the way to go and adding the static serve is helpful in debugging and not harmful in production. Config wise using the compose volume wps_outputs and mounting it in both docker image at a proper place should do it.

davidcaron:
aims2 is still down, I think it's been close to 2 weeks now...

MatProv:
This pull request will not be merge.
See `esgf-cwt-process-rebase`, for an updated version.
Closing this old PR.

PR #26, opened 2019-03-26
try to append a slash to the path when the route is not found

davidcaron:


davidcaron:
@fmigneault This is not WIP. I should have added a description, here is a link to the documentation: https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/urldispatch.html#redirecting-to-slash-appended-routes

The goal is to redirect routes like `/providers` to `/providers/`. Currently, `/providers/` returns a 404, while `/providers` works. Using this configuration, both formats are good.

fmigneault:
I marked it as WIP only because not tested and as conflict.
I don't mind to have it if tests all pass.

PR #40, opened 2019-05-06
Fix email config

davidcaron:
Some fixes for the email configuration were not set properly in buildout. And raise the proper exception message when trying to send an email with a wrong configuration.


davidcaron:
@fmigneault I'm not sure I'm doing it right with buildout... should I add these settings in buildout.cfg?

fmigneault:
@davidcaron yes, you're doing it correctly.

PR #42, opened 2019-05-14
encode notification emails in utf8

davidcaron:


weaver/wps_restapi/templates/notification_email_example.mako
```diff
@@ -1,3 +1,4 @@
+## -*- coding: utf-8 -*-
```
fmigneault: Does the double # make a difference?

davidcaron: I copied from the documentation... But something as simple as ‘#coding: utf8’ would work I think.

Issue #11, opened 2019-03-09
functional tests for process deployment


fmigneault:
relates to #3

- [x] add test to deploy process with CWL `DockerRequirement`
    - [x] reference .cwl in `href` instead of `ProcessDescription` (**see `DockerRequirement` note**)
    - [x] reference .cwl in `owsContext` within `ProcessDescription -> Process`
    - [x] reference .cwl in `executionUnit` (combined with `deploymentProfileName`) (**see `DockerRequirement` note**)
- [x] add test to deploy process with CWL `WPS1Requirement` (added in #372)
  (must provide `process` in requirement)
    - [x] reference .cwl in `href` instead of `ProcessDescription`
    - [x] reference .cwl in `owsContext` within `ProcessDescription -> Process` 
    - [x] reference .cwl in `executionUnit` (combined with `deploymentProfileName`)
- [x] add test to deploy process with WPS1 `DescribeProcess` endpoint
    - [x] reference WPS1 in `href` instead of `ProcessDescription`
    - [x] reference WPS1 in `owsContext` within `ProcessDescription -> Process`
    - [x] reference WPS1 in `executionUnit` (combined with `deploymentProfileName`)
- [x] add test to deploy process with WPS1 service `GetCapabilities` endpoint 
  (must include `identifier=...` to derive `DescribeProcess`)
    - [x] reference WPS1 in `href` instead of `ProcessDescription`
    - [x] reference WPS1 in `owsContext` within `ProcessDescription -> Process`
    - [x] reference WPS1 in `executionUnit` (combined with `deploymentProfileName`)
- [x] add test to deploy process with WPS3 (REST) `GET Process` endpoint
    - [x] reference WPS3 in `href` instead of `ProcessDescription`
    - [x] reference WPS3 in `owsContext` within `ProcessDescription -> Process`
    - [x] reference WPS3 in `executionUnit` (combined with `deploymentProfileName`)
- [x] #434
    - [x] package directly provided as JSON
    - [x] reference to remote CWL package
- [x] test with `Workflow` using deployed sub-processes

### `DockerRequirement` note

Implementation missing from below tests  but feature is tested across many other tests since it is the most common use case.
Should implement just to make sure it is always evaluated for consistency.
https://github.com/crim-ca/weaver/blob/107fec5e19f20b77061b9405a764da911d2db8a2/tests/wps_restapi/test_processes.py#L627-L641

dbyrns:
[...]

PR #46, opened 2019-05-31
Job email and grouped results for stats

### LINKED ISSUE #44: Save notification email hash of job submission
- On posted job, generate a hash of the email (to make them anonymous) saved to the db.
- Allow search filter of job by email hash.

fmigneault:
- add `Job` email encrypted field 
- search jobs by email (hashed)
- group jobs by field categories

Resolves #44 

## todo
- complete #48


fmigneault:
failing tests are related to #11 being implemented/fixed in https://github.com/crim-ca/weaver/tree/func-tests

weaver/store/mongodb.py
```diff
@@ -477,6 +526,9 @@ def find_jobs(self,
         elif status:
             search_filters["status"] = status
 
+        if notification_email is not None:
+            search_filters["notification_email"] = notification_email
```
dbyrns: I guess that we should search the encrypted mail and not the real mail?

fmigneault: Done [here](https://github.com/crim-ca/weaver/pull/46/files/06a87735547afbf6a390868999e7d8f45557c4bf#diff-983083985a48e53e4a69189e6dbe2c53R207) and validated by [this test](https://github.com/crim-ca/weaver/pull/46/files/06a87735547afbf6a390868999e7d8f45557c4bf#diff-19c25203fab87905fceae3782f9a36ecR308)

fmigneault: @tomLandry 
To filter by email, you have to specify the literal email to match against their hashed version in the db.
You can't search with the hashed version. 
Unless you have the secret & salt values to know how to pass from one-another, you can't find which hash corresponds to which email literal value, so identifies masking is preserved.

Furthermore, email values are not returned in responses. 
So you can only go by trial-error of literal email values to attempt matching jobs, and if you happen to match jobs this way, then you already have the email to create the request in the first place.

PR #77, opened 2020-02-11
fix for accept_language when empty or None

davidcaron:
- Could be None when updating weaver
 - The serialized string must be empty when no accept-language header is provided


davidcaron:
CI fails for the same reasons as on master, merging this.

PR #81, opened 2020-02-19
Don't cast values when validating a job's POST body

davidcaron:
For example, an input of `{"id": "test", "data": "2020"}`
was cast to `{"id": "test", "data": 2020.0}`
because `DataFloat` is the first item in `weaver.wps_restapi.swagger_definitions.ValueType`

### DIFF
--- a/weaver/wps_restapi/processes/processes.py
+++ b/weaver/wps_restapi/processes/processes.py
@@ -315,7 +315,7 @@ def submit_job_handler(request, service_url, is_workflow=False, visibility=None)
     except Exception as ex:
         raise HTTPBadRequest("Invalid JSON body cannot be decoded for job submission. [{}]".format(ex))
     try:
-        sd.Execute().deserialize(json_body)
+        json_body = sd.Execute().deserialize(json_body)
     except colander.Invalid as ex:
         raise HTTPBadRequest("Invalid schema: [{}]".format(str(ex)))
 
--- a/weaver/wps_restapi/processes/processes.py
+++ b/weaver/wps_restapi/processes/processes.py
@@ -315,7 +315,7 @@ def submit_job_handler(request, service_url, is_workflow=False, visibility=None)
     except Exception as ex:
         raise HTTPBadRequest("Invalid JSON body cannot be decoded for job submission. [{}]".format(ex))
     try:
-        json_body = sd.Execute().deserialize(json_body)
+        sd.Execute().deserialize(json_body)
     except colander.Invalid as ex:
         raise HTTPBadRequest("Invalid schema: [{}]".format(str(ex)))
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,13 @@ Changes
 Unreleased
 ==========
 
+Fixes:
+------
+
+- Fix a bug where the validation of OneOf items was casting the value to the first valid possibility.
+  Now, it doesn't change the value if it's valid without casting it (and still casts it if it's 
+  necessary to make it valid).
+
 1.1.0 (2020-02-17)
 ==================
 
--- a/tests/wps_restapi/test_processes.py
+++ b/tests/wps_restapi/test_processes.py
@@ -544,6 +544,28 @@ class WpsRestApiProcessesTest(unittest.TestCase):
             assert resp.status_code in [400, 422], msg.format(i, resp.status_code)
             assert resp.content_type == CONTENT_TYPE_APP_JSON, msg.format(i, resp.content_type)
 
+    def test_execute_process_dont_cast_one_of(self):
+        """
+        Optional parameters for execute job shouldn't raise an error if omitted,
+        and should resolve to default values if any was specified.
+        """
+        # get basic mock/data templates
+        name = fully_qualified_name(self)
+        execute_mock_data_tests = list()
+
+        mock_execute = mocked_process_job_runner("job-{}".format(name))
+        data_execute = self.get_process_execute_template("100")
+        execute_mock_data_tests.append((mock_execute, data_execute))
+
+        with contextlib.ExitStack() as stack:
+            for exe in mock_execute:
+                stack.enter_context(exe)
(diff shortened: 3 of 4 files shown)


weaver/wps_restapi/processes/processes.py
```diff
@@ -315,7 +315,7 @@ def submit_job_handler(request, service_url, is_workflow=False, visibility=None)
     except Exception as ex:
         raise HTTPBadRequest("Invalid JSON body cannot be decoded for job submission. [{}]".format(ex))
     try:
-        json_body = sd.Execute().deserialize(json_body)
+        sd.Execute().deserialize(json_body)
```
fmigneault: This causes optional values from the input, but required later below, not to be filled in the body with defaults.

fmigneault:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
- Added 2
           

Complexity increasing per file
==============================
- weaver/wps_restapi/colander_extras.py  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/fmigneault/weaver/pullRequest?prid=5033780&bid=16582968)

davidcaron:
@fmigneault I think this should work... I compare the valid schemas with the source, and if one of them is equal to the source, I return it. 

If that's not possible, I return the first valid schema (just like it was before this PR).

fmigneault:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- weaver/wps_restapi/colander_extras.py  3
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/fmigneault/weaver/pullRequest?prid=5033780&bid=16582968)

fmigneault:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- weaver/wps_restapi/colander_extras.py  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/fmigneault/weaver/pullRequest?prid=5033780&bid=16582968)

davidcaron:
@fmigneault Tests pass now.

fmigneault:
The is one use case which I'm still concerned with before integrating this.
Its related to min/maxOccurs parameters that can be provided as string/float. 
Because "unbounded" can be used, string is considered the valid type (and the expected one by many WPS servers).

All tests in https://github.com/crim-ca/weaver/blob/master/tests/functional/test_wps_package.py validate and use string values. I would be interested to see if explicitly providing a valid int value would be auto-resolved to string or left as int. Maybe the missing cast will break things lower down the execution chain.

davidcaron:
If the user provides it as an int, it should stay an int with this PR.

It was converted to a string before, so the server would receive `"2"` instead of `2`. The server logic probably already implemented a `try...except` to convert the data to an int.

In my case, the string was converted to a float, so `"2000"` was becoming `2000.0` so it would have been very awkward to implement the logic to get back a string in the pywps process, when the process explicitly asked for a string.

Maybe you want to add a test?

fmigneault:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- weaver/wps_restapi/colander_extras.py  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/fmigneault/weaver/pullRequest?prid=5033780&bid=16582968)

fmigneault:
> It was converted to a string before, so the server would receive `"2"` instead of `2`. The server logic probably already implemented a `try...except` to convert the data to an int.

Yes, it's this part that I fear. Weaver is not the only implementation of ADES/EMS. 
I wouldn't be surprised that other servers only take the string value (the remote ones where the requests are forwarded during workflows).

This is the only implementation that is more permissive to allow the numeric equivalent value, but it has to be converted to string to support "unbounded" and to respect the *potentially* string only value on remotes.

For your use case, I agree that the input data should not be cast'd.

davidcaron:
From the reference implementation, `minOccurs` must be an integer, and `maxOccurs` must be either an integer or the string `"unbounded"`. So an integer inside a string is not allowed according to this. https://github.com/opengeospatial/wps-rest-binding/blob/master/core/openapi/schemas/inputDescription.yaml

The 52North demo uses integers: http://geoprocessing.demo.52north.org:8080/javaps/rest/processes/org.n52.wps.server.algorithm.SimpleBufferAlgorithm

The pygeoapi demo uses integers also: https://demo.pygeoapi.io/master/processes/?f=json

Right now, in weaver, it's converted to a string no matter what, so it seems to me that weaver does not conform to the standard.

fmigneault:
Don't get me wrong, I agree and personally would prefer to have only integers. 
The issue is we need to support cases where WPS-1 define XML attributes as strings (ex: 
http://colibri.crim.ca:8093/wps?service=WPS&request=describeprocess&version=1.0.0&identifier=subset_bbox)
So if the conversion happens correctly when parsing a remote WPS-1 process, we can apply this change.

davidcaron:
I have a docker image of weaver deployed that uses this PR at: https://finch.crim.ca/providers/finch/processes/subset_bbox

Does this help in figuring out if this change breaks anything?

fmigneault:
Seems to be handled correctly for the min/max as the resulting values are left as strings...

I see many more internal parameters. Is it caused by this change ? 
https://finch.crim.ca/providers/finch/processes/ice_days
vs
https://ogc-ems.crim.ca/weaver/processes/Finch_IceDays

davidcaron:
> https://finch.crim.ca/providers/finch/processes/ice_days

No, it's only a newer version of finch.

fmigneault:
OK. 
Please add a changelog entry and we can merge.

fmigneault:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- weaver/wps_restapi/colander_extras.py  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/fmigneault/weaver/pullRequest?prid=5033780&bid=16582968)

Issue #78, opened 2020-02-11
Update owslib to for Accept-Language header


fmigneault:
update package requirement whenever available (https://github.com/geopython/OWSLib 0.20.0)
see https://github.com/crim-ca/weaver/pull/74

PR #89, opened 2020-03-13
update owslib to 0.19.2

davidcaron:


davidcaron:
@fmigneault any minor version release planned soon?

fmigneault:
I was waiting for the package (#78)
Can be udpated soon, but need to fix the requirements. 
Python 3.5 fails because owslib dropped it.

davidcaron:
If owslib dropped support for python 3.5, you're right that this would mean weaver should to when we update to this version.

setup.py
```diff
@@ -25,7 +25,6 @@
       classifiers=[
           "Natural Language :: English",
           "Programming Language :: Python :: 2.7",
-          "Programming Language :: Python :: 3.5",
```
fmigneault: also update the below `python_requires`

davidcaron: done

PR #128, opened 2020-05-07
retry reuse + mime-type handling

fmigneault:
- improve and reuse the common request/retry function to handle potential temp failures
- improve mime-type parsing to consider extra parameters and extended supported ones


fmigneault:
@MatProv 
Don't know why I can't add you as reviewer :/ 
But this is the fix about the content-type

Issue #25, opened 2019-03-25
How to handle "multiple output"


fmigneault:
According to `owslib.wps`, `minOccurs`/`maxOccurs` can only be parsed for inputs, outputs are always a single reference or literal data.

In `pywps`, the outputs have a reference to `min_occurs`/`max_occurs` because of inheritance from base classes. They are also returned in their `json` property.

For `CWL`, there is no restriction to have multiple outputs (array of string, file, etc.).
Actually, when using a regex `glob: *.<ext>` pattern, a list of results is automatically created with all matches.

How can we produce the `CWL <-> WPS` conversion in these cases?
Also, should we enforce explicit values of `minOccurs`/`maxOccurs` : 
  -  ~`minOccurs="1"` enforced since there should always be an output~
     **update**: *since CWL allows optional outputs (eg: `File?`), `minOccurs=0|1` should be set according to nullable detection*
  -  `maxOccurs="1" | "<number>" | "unbounded"`, with `"1"` being the default (same reason as min), or one of `"<number>" | "unbounded"` if specified during deployment

ref #17 
relates to https://github.com/opengeospatial/wps-rest-binding/issues/37

# To Do 

- [ ] check any test with commented references to https://github.com/crim-ca/weaver/issues/25

tomLandry:
Unsure if that answers your question, but please take a look at metalinks: https://github.com/bird-house/emu/issues/64. First implementation here: https://github.com/bird-house/emu/commit/1d6125f89fbb5a6dea3487edb53bbaccb1c43ec8

This is a recommendation from ADES&EMS best practices ER: https://docs.opengeospatial.org/per/18-050r1.html#_multiple_outputs

fmigneault:
see method using metalink like in birdy:
https://github.com/bird-house/birdy/blob/26798f62c6d1add6d28acd8e283c5f7d05083ffd/birdy/client/converters.py#L99-L115

fmigneault:
implementation in pywps: 
https://github.com/geopython/pywps/issues/298

tomLandry:
@fmigneault Multiple outputs is an "evaluation aspect" of the OGC EO Pilot project. Thus we need to prioritize this relatively high. See https://portal.opengeospatial.org/files/?artifact_id=90733#_application_execution_multiple_outputs

fmigneault:
Via pypws : 
https://github.com/geopython/pywps/blob/master/docs/metalinkprocess.py

fmigneault:
[...]

PR #132, opened 2020-05-12
Json output + job captured logging

fmigneault:
- Fixes incorrect parsing of JSON output as 'multiple-output' because of new job location that now stores a relative path to file (relates to #25)
- Add more execution log in to saved job logs adding the captured stderr & stdout of the underlying docker app or script.

comparison examples:  
[no-capture.log](https://github.com/crim-ca/weaver/files/4613378/no-capture.log)
[with-capture.log](https://github.com/crim-ca/weaver/files/4613382/with-capture.log)

### DIFF
--- a/weaver/processes/utils.py
+++ b/weaver/processes/utils.py
@@ -1,5 +1,6 @@
 import json
 import logging
+import os
 import warnings
 from copy import deepcopy
 from distutils.version import LooseVersion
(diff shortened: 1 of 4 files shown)


weaver/processes/utils.py
```diff
@@ -84,29 +86,42 @@ def _read_reference(url):
...
+    # Check for the json datatype and mime-type
     if output.dataType == WPS_COMPLEX_DATA and output.mimeType == CONTENT_TYPE_APP_JSON:
+        try:
+            # If the json data is referenced read it's content
+            if output.reference:
+                out_ref = output.reference
+                if container:
+                    if out_ref.startswith("file://"):
+                        out_ref = out_ref[7:]
+                    if out_ref.startswith("/"):
```
dbyrns, line 111: absolute file format used as relative path? Why not starting with a "."?

fmigneault: URL path, not directory path
for convenience, files are saved in db as /job-id.... so that we just add wps-outputs path in front regardless of (possibly changing) setting

PR #147, opened 2020-05-19
patches

fmigneault:
- pserve in docker  (to correctly paste ini workers and host/port)
- requests-extra function to have per-url/method extra arguments (SSL error handling in specific cases)
- WPS I/O Literal Data type fix when retrieved from owslib object which stored directly the type instead of within other `data_type` field.


fmigneault:
@MatProv you will need these fixes

Issue #155, opened 2020-05-24
[Feature] adjust user/group id for running docker 


fmigneault:
Using `fixuid`, we could potentially solve the issues related to user permissions. 
We need to enforce a `worker` user so that the docker is not executed with elevated `root` permissions. This is *required*  if we want to do sibling docker containers with minimal security/control.

references: 
- https://boxboat.com/2017/07/25/fixuid-change-docker-container-uid-gid/
- https://github.com/boxboat/fixuid
- https://github.com/boxboat/fixuid/blob/master/docker/debian/Dockerfile

Also, we need to enforce `no_read_only=False` such that files are mounted as read-only. This was workaround previously, but the correct way to handle modifiable files from the app's workdir is to use `InitialWorkDirRequirement` (ex: https://www.commonwl.org/user_guide/15-staging/) so that files are moved with write permissions as needed in the app.

https://github.com/crim-ca/weaver/blob/ede8098736185e87d5c608647cd9a59249739049/weaver/processes/wps_package.py#L1966-L1967

PR #162, opened 2020-05-29
config workdir and type handling

fmigneault:
- extend features of PR #150 using `weaver.wps_workdir` 
    - better setup cwltool tmpdir and results location
    - functional with docker-sibling setup
- fixes incorrect reference when default type parsing is used (introduced by PR #128)
- addresses #155 [experimental] with user/group applied to cwl app & readonly inputs
- avoid reload pywps configs every time
- better setup of cwltool logs according to the rest of the app log level
- map wps-outputs with job uuid to pywps uuid to make it easier to identify matching items

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,22 @@ Changes
 `Unreleased <https://github.com/crim-ca/weaver/tree/master>`_ (latest)
 ========================================================================
 
+Changes:
+--------
+
+- Add ``weaver.wps_workdir`` configuration setting to define the location where the underlying ``cwltool`` application
+  should be executed under.
+- Use ``weaver.request_options`` for `WPS GetCapabilities` and `WPS Check Status` requests under the running job.
+- Change default ``DOCKER_REPO`` value defined in ``Makefile`` to point to reference mentioned in ``README.md`` and
+  considered as official deployment location.
+
+Fixes:
+------
+
+- Set ``get_cwl_file_format`` default argument ``must_exist=True`` instead of ``False`` to retrieve original default
+  behaviour of the function. Since `CWL` usually doesn't need to add ``File.format`` field when no corresponding
+  reference actually exists, this default also makes more sense.
+
 `1.8.1 <https://github.com/crim-ca/weaver/tree/1.8.1>`_ (2020-05-22)
 ========================================================================
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,8 +8,13 @@ Changes:
 --------
 
 - Add ``weaver.wps_workdir`` configuration setting to define the location where the underlying ``cwltool`` application
-  should be executed under.
+  should be executed under. This can allow more control over the scope of the mounted volumes for *Application Package*
+  running a docker image.
+- Add *experimental* configuration settings ``weaver.cwl_euid`` and ``weaver.cwl_egid`` to provide effective user/group
+  identifiers to employ when running the CWL *Application Package*. Using these require good control of the directory
+  and process I/O locations as invalid permissions could break a previously working job execution.
 - Use ``weaver.request_options`` for `WPS GetCapabilities` and `WPS Check Status` requests under the running job.
+- Enforced removal of some invalid `CWL` hints/requirements that would break the behaviour offered by ``Weaver``.
 - Change default ``DOCKER_REPO`` value defined in ``Makefile`` to point to reference mentioned in ``README.md`` and
   considered as official deployment location.
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,8 +13,16 @@ Changes:
 - Add *experimental* configuration settings ``weaver.cwl_euid`` and ``weaver.cwl_egid`` to provide effective user/group
   identifiers to employ when running the CWL *Application Package*. Using these require good control of the directory
   and process I/O locations as invalid permissions could break a previously working job execution.
+- Add more logging configuration and apply them to ``cwltool`` before execution of *Application Package*.
+- Enforce ``no_match_user=False`` and ``no_read_only=False`` of ``cwltool``'s ``RuntimeContext`` to ensure that docker
+  application is executed with same user as ``weaver`` and that process input files are not modified inplace (readonly)
+  where potentially inaccessible (according to settings). Definition of `CWL` package will need to add
+  `InitialWorkDirRequirement <https://www.commonwl.org/v1.0/CommandLineTool.html#InitialWorkDirRequirement>`_ as per
+  defined by reference specification to stage those files if they need to be accessed with write permissions
+  (see: `example <https://www.commonwl.org/user_guide/15-staging/>`_). Addresses some issues listed in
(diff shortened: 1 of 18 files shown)


weaver/processes/wps_package.py
```diff
@@ -1962,9 +1979,34 @@ def _handler(self, request, response):
...
-            wps_out_dir_prefix = os.path.join(get_wps_output_dir(settings), "tmp")
-            runtime_args = {"no_read_only": True, "outdir": self.workdir, "tmp_outdir_prefix": wps_out_dir_prefix}
-            runtime_context = RuntimeContext(kwargs=runtime_args)
+            self.update_effective_user()
+            self.update_requirements()
+
+            # note:
+            #   Parameter 'weaver.wps_workdir' is the base-dir where sub-dir per application packages will be generated.
+            #   Parameter 'self.workdir' is the actual location PyWPS reserved for this process (already with sub-dir).
+            #   If no 'weaver.workdir' was provided, reuse PyWps parent workdir since we got access to it.
```
dbyrns, line 1988: Since it's already a little bit confusing, we should make sure that there is no error in doc. I guess here we should read weaver.wps_workdir and not weaver.workdir?

dbyrns, line 1988: If we reuse PyWps parent workdir, it means that cwl app packages private dir will exist along proces dir? We will then expose those private directories on the web? 
Why not requiring such a parameter or creating one under /tmp ?

weaver/wps_restapi/jobs/jobs.py
```diff
@@ -38,7 +38,30 @@
...
+    :param must_exist: return only existing path if enabled, otherwise return the parsed value without validation.
+    :returns: found local file path if it exists, ``None`` otherwise.
+    """
+    if url_status_location and not urlparse(url_status_location).scheme in ["", "file://"]:
+        dir_path = get_wps_output_dir(container)
+        wps_out_url = get_wps_output_url(container)
+        req_out_url = get_url_without_query(url_status_location)
+        out_path = os.path.join(dir_path, req_out_url.replace(wps_out_url, "").lstrip("/"))
+    else:
+        out_path = url_status_location.replace("file:://", "")
```
dbyrns, line 56: typo in file::// (double ::) or this is really a case?

weaver/wps_restapi/processes/processes.py
```diff
@@ -216,11 +219,13 @@ def execute_process(self, job_id, url, headers=None, notification_email=None):
                         job.status_message = "Job failed{}.".format(msg_progress)
                         wps_package.retrieve_package_job_log(execution, job)
                         job.save_log(errors=execution.errors, logger=task_logger)
+                    task_logger.debug("Mapping Job references with generated WPS locations.")
+                    map_locations(job, settings)
```
dbyrns, line 223: You map what here? The current job is monitored by PyWPS, so, if the job run a local PyWPS process (only case I see is for workflows?) you link directly the job process status to the api process status?

CHANGES.rst
```diff
@@ -4,6 +4,40 @@ Changes
...
+
+- Add ``weaver.wps_workdir`` configuration setting to define the location where the underlying ``cwltool`` application
+  should be executed under. This can allow more control over the scope of the mounted volumes for *Application Package*
+  running a docker image.
+- Add mapping of WPS results from the ``Job``'s UUID to generated `PyWPS` UUID for outputs, status and log locations.
+- Add *experimental* configuration settings ``weaver.cwl_euid`` and ``weaver.cwl_egid`` to provide effective user/group
+  identifiers to employ when running the CWL *Application Package*. Using these require good control of the directory
+  and process I/O locations as invalid permissions could break a previously working job execution.
+- Add more logging configuration and apply them to ``cwltool`` before execution of *Application Package*.
+- Enforce ``no_match_user=False`` and ``no_read_only=False`` of ``cwltool``'s ``RuntimeContext`` to ensure that docker
```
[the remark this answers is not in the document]
fmigneault: Both `no_match_user` and `no_read_only` are now *enforced* using the above values, so input files are readonly and user is mapped. Applications that expect non-readonly files can use the staging method using `InitialWorkDirRequirement` which is the standard CWL way of doing it. This makes the file to be loaded in the runtime dir, where permissions are writable to allow dumping the process outputs.
 
Only issue that I notice while testing; since `weaver` is executed with docker-compose with admin-level user, the mapped user is `0:0`. So unless we override the user ids with some 'worker' to run everything, the apps are still executed as root...

weaver/processes/wps_package.py
fmigneault: Indeed. Fixing it.

fmigneault: The pywps/cwltool directories are at the same level, but the workdir is not on the wps-outputs path. It is some `/tmp/whatever/pywps_process_<random>` location with `/tmp/whatever` configured using `weaver.wps_workdir`. The results generated by this are automatically moved to path corresponding to `{weaver.wps_output_dir}/{job-id}` when the process completes.

weaver/wps_restapi/jobs/jobs.py
fmigneault: Intentional thing that I found a while ago. I don't recall 100% the use case... should have commented it back then. I think is was related to pywps doing this modification. Maybe it is not the case anymore, I have not seen this else branch being called during my recent tests.

weaver/wps_restapi/processes/processes.py
fmigneault: In the wps-outputs, the job-id, and pywps 'process-execution-id' don't match. 
Technically, there is not problem, you query the API route `/jobs/{jobId}/results`, and each path in the outputs have `{wps-outputs-path}/{processExecId}/{the-output}`. It is the same of the xml status and log files generated. They have the name based on `{processExecId}`. 
The map only attempts to create a symlink such that `{jobId}->{processExecId}`. 
This is not a 'functional' requirement, but more a 'for convenience' one. 
Avoids me going through hundreds of job logs to find the other ID to do the mapping manually.

fmigneault: Example
image

fmigneault:
@MatProv  @dbyrns  
I got the tests passing with last commit, so let me know if there are any other issues/comments.
I plan to merge in a bit so I can build and try it on some server.

CHANGES.rst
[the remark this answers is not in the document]
dbyrns: So since our applications expect non-readonly files, you confirm that with `InitialWorkDirRequirement` they are still working?
For the root issue, that's why you introduced the user id, group id settings? Maybe you could use a string and use the gpt-init script trick to extract their values at runtime?

weaver/wps_restapi/processes/processes.py
dbyrns: Clever, I remember digging into that stuff!

CHANGES.rst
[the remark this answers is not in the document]
dbyrns: Got a call, everything is fine! Will postpone the user/group handling because the solution should be to run weaver as a limited user rather than root.

PR #166, opened 2020-06-04
Remove stdout stderr workflow

MatProv:
- Avoid logging ``stdout/stderr`` in workflows
- Add tests to make sure processes ``stdout/stderr`` are logged


CHANGES.rst
```diff
@@ -30,6 +30,8 @@ Changes:
 - Add ``application/x-yaml``  MIME-type to known formats.
 - Add ``application/x-tar`` and ``application/tar+gzip`` MIME-type (not official) but resolved as *synonym*
   ``application/gzip`` (official) to preserve compressed file support during `CWL` format validation.
+- Avoid logging ``stdout/stderr`` in workflows
+- Add tests to make sure processes ``stdout/stderr`` are logged
```
fmigneault, line 34: move to top with latest changes

tests/processes/echo.sh
```diff
@@ -0,0 +1 @@
+echo $1
```
fmigneault, line 1: newline

tests/processes/test_wps_package.py
```diff
@@ -494,3 +500,155 @@ def test_merge_io_formats_wps_overlaps_cwl():
...
+                 }
+            ]
+        },
+        "outputs":{
+
+        }
+    }
+
+    # ExecuteResponse mock
+    wps_response = type('',(object,),{"_update_status": lambda w,x,y,z: 1 })()
```
fmigneault, line 565: `lambda *_, **__: 1` to avoid problem with any future changes in signature

tests/processes/test_wps_package.py
```diff
@@ -494,3 +500,155 @@ def test_merge_io_formats_wps_overlaps_cwl():
         Format(CONTENT_TYPE_APP_XML),
         Format(CONTENT_TYPE_TEXT_PLAIN),
     ])
+
+
+def test_stdout_stderr_logging_for_commandline_tool_success():
```
fmigneault, line 505: add a docstring that says the evaluated criteria of the test so we can understand the purpose at a glance

tests/processes/test_wps_package.py
```diff
@@ -494,3 +500,155 @@ def test_merge_io_formats_wps_overlaps_cwl():
...
+    wps_package_instance.workdir = workdir.name
+    wps_package_instance._handler(wps_request, wps_response)
+
+    # log assertions
+    with open(status_location + ".log", "r") as file:
+        log_data = file.read()
+        assert "Dummy message" in log_data
+
+
+def test_stdout_stderr_logging_for_commandline_tool_failure():
```
fmigneault, line 581: criteria docstring

fmigneault: Also, we could make a common function that setups all the boilerplate code, and only replace the `not_existing_command` vs `echo` in each case to remove duplication.

weaver/processes/wps_package.py
```diff
@@ -2107,7 +2112,8 @@ def _handler(self, request, response):
             raise
         else:
             self.update_status("Package complete.", PACKAGE_PROGRESS_DONE, STATUS_SUCCEEDED)
-        return self.response
+        finally:
+            return self.response
```
fmigneault, line 2116: why the finally here?

tests/functional/test_ems_end2end.py
```diff
@@ -421,6 +421,9 @@ def log_json_format(cls, payload, indent_level):
     @classmethod
     def log_dict_format(cls, dictionary, indent_level):
         """Logs dictionary (key, value) pairs in a YAML-like format."""
+        if dictionary is None:
+            return ""
```
fmigneault, line 425: Return `<None>` to make it explicit in the log that container was empty.

tests/processes/test_wps_package.py
MatProv: Good idea, this amelioration would be done at the same time we add the stdout/stderr for the workflows

weaver/processes/wps_package.py
MatProv: In order to let the handler finish even if it raises an exception. Otherwise, lets say we have a process which fails, "Completed PermanentFailed" is returned but the the execution is stopped here. We want the error to be sent to logs but not to hang the execution since there was an issue.

MatProv:
@fmigneault there are no linting recommendation

fmigneault:
@MatProv 
> 
> there are no linting recommendation

You can find it through the `Travis CI` action of the matching `X` of each commit.
For 67d78bc, they are here https://travis-ci.com/github/crim-ca/weaver/jobs/344358776

weaver/processes/wps_package.py
fmigneault: Will re-evaluate when travis re-runs everything after linting is fixed. 
Not sure if not raising the exception correctly updates the response as failed with the traceback for debuging...

MatProv: Yes, you're right. I got this after linting : "return statement in finally block may swallow exception (lost-exception)". I however don't have a proposition, since I think we want handler to raise, but correctly finish, to avoid hanging up the execution..

fmigneault: I think that since each raise includes a status_update call beforehand with the appropriate message and job-status value, there is no need for anything else. PyWPS then catches the exceptions and logs everything as needed.

tests/opensearch_examples/OpenSearch_Collection_Search.xml
```diff
@@ -1,4 +1,8 @@
-<?xml version="1.0" encoding="UTF-8"?><feed xmlns="http://www.w3.org/2005/Atom" xmlns:csw="http://www.opengis.net/cat/csw/2.0.2" xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:fn="http://www.w3.org/2005/xpath-functions" xmlns:geo="http://a9.com/-/opensearch/extensions/geo/1.0/" xmlns:georss="http://www.georss.org/georss" xmlns:gml="http://www.opengis.net/gml" xmlns:os="http://a9.com/-/spec/opensearch/1.1/" xmlns:rim="urn:oasis:names:tc:ebxml-regrep:xsd:rim:3.0" xmlns:semantic="http://a9.com/-/opensearch/extensions/semantic/1.0/" xmlns:sru="http://a9.com/-/opensearch/extensions/sru/2.0/" xmlns:time="http://a9.com/-/opensearch/extensions/time/1.0/" xmlns:wrs="http://www.opengis.net/cat/wrs/1.0" xmlns:xlink="http://www.w3.org/1999/xlink">
```
fmigneault, line 1: Why are the references removed?

MatProv: removed by fix-imports

fmigneault: I just ran `make fix-imports` and files are not modified.
They shouldn't also, they are not even python files... not sure what really happened there.

MatProv: Not sure what happened too but the four xml files are now reverted.

PR #170, opened 2020-06-09
more docs

fmigneault:
- another batch of documentation updates
- fix some links/references to generate doc OpenAPI


docs/source/configuration.rst
```diff
@@ -14,11 +15,135 @@ After you have installed `Weaver`, you can customize its behaviour using multipl
...
+  | ``weaver.wps_output_url = <full-url>``
+  | (default: *path* ``/wpsoutputs``)
+  |
+  | Endpoint that will be employed as prefix to refer to WPS outputs (job results).
+  |
+  | It can either be the explicit *full URL* to use or the *path* relative to ``weaver.url``.
+  | Setting ``weaver.wps_output_path`` is ignored if its URL equivalent is defined.
+  | The *path* variant **SHOULD** start with ``/`` for appropriate concatenation with ``weaver.url``, although this is
+    not strictly enforced.
+
```
MatProv, line 88: Would it be better to enforce it, instead of having runtime issues if this detail is missing? Either by giving a warning when config is getting read or by adding the '/' if absent.

fmigneault: Warning could be a good idea, but since this setting is read really often, I don't want to spam too much logs either. We could need to have smart warning that disables itself after the first occurrence. 

We shouldn't enforce since the path without `/` is valid, it really depends on how `weaver.url` is defined. For example, if someone prefers to use `weaver.url` finishing with `/`, enforcing one in `weaver.wps_output_path` would add it twice.

MatProv: I see. Therefore adding or removing the / could be during the merge or url/path. Not a priority for now but a simple detail that can create error/confusion as the platform grows and more technicalities about urls are defined (eg: this one is / prefixed, the other one / suffixed, this url path no slash, this one prefixed and suffixed).. If we have 3-4 urls/paths well I think that's ok but as the config needs grows, it could be helpful to handle those cases.

fmigneault: I prefer to avoid as much as possible to add/remove anything inplace. If the config is invalid, the user should simply update it to make is correct. When the config is invalid, usually you can see it rather quickly as something is not working as expected.

I have had too many situations when an extra or missing `/` caused me so many problems within nginx configurations. I'd personally rather know that when I explicitly set a `/` at some location (regardless where it is), it is applied as is and the code doesn't "help me" undo my configuration.

MatProv: Good point, then specifying the url format in the doc as you did is perfect.

PR #176, opened 2020-06-18
Esgf cwt process rebase

MatProv:
- Modifications to #23 in order to be compatible with the weaver version of today
- Adds builtins and workflows

Notes:
- The `config.add_static_view` discussion on #23 will be part of an issue, to keep traces
- Branches `esgf-cwt-process` and `esgf-cwt-process` will be closed once this one is merged

PR #181, opened 2020-07-02
add minor version prior to esgf merge

fmigneault:
@MatProv 
I find the changes introduced in #176 to be a bit too ambitious to be merged as is into master with already many unversioned changes.  So I pushed a `1.11.0` version at the commit just prior to all the ones from #176. 

I also find that such massive change are still quite lacking in term of changelog.
Can you please summarize all changes applied within that PR under the `Unreleased` section?
After this, we will immediately add another tag `1.12.0`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,17 @@ Changes
 
 Changes:
 --------
+- No change.
+
+Fixes:
+------
+- No change.
+
+`1.11.0 <https://github.com/crim-ca/weaver/tree/1.11.0>`_ (2020-07-02)
+========================================================================
+
+Changes:
+--------
 - Generate Weaver OpenAPI specification for readthedocs publication.
 - Add some sections for documentation (`#61 <https://github.com/crim-ca/weaver/issues/61>`_).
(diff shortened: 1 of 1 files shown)


MatProv:
@fmigneault Of course, changelog has been forgotten in #176 so it should be updated quickly before we forget the changes.
I've added the changes. Therefore i'm not sure of the bump procedure (only `make bump 1.12.0` + `git push --tags` ?), so I haven't pushed a new tag. Let me know if this is right, so I can do it and then merge.

fmigneault:
> 
> 
> @fmigneault Of course, changelog has been forgotten in #176 so it should be updated quickly before we forget the changes.
> I've added the changes. Therefore i'm not sure of the bump procedure (only `make bump 1.12.0` + `git push --tags` ?), so I haven't pushed a new tag. Let me know if this is right, so I can do it and then merge.

I would move the 
> Reset ``MongoDatabase`` connection when we are in a forked process.

to the "fixes" as it patches the warning logged everytime. 

Then, after this is merged, I can run a `make VERSION=1.12.0 bump` and push the new tag.
I think I will have to do it as it requires push to master.

MatProv:
@fmigneault alright, changelog updated

Issue #183, opened 2020-07-10
[Feature] Explicit 'fetch' parameter for some given input


fmigneault:
Weaver does a lot of implicit decisions about downloading a file or not in order to try reducing the amount of unnecessary downloads. 

For example, WPS-1 processes monitored by Weaver don't require fetching the file on Weaver side since it will be pulled by the WPS anyway. Same goes for Weaver EMS that dispatches the execution to another ADES.

There are some cases where this behavior could be undesired, or that resolution is ambiguous. 
For example: 
- should an S3 bucket reference be resolved by Weaver EMS to provide it as HTTP to some ADES that cannot access the S3 directly, or should the reference be directly forwarded to the ADES which is the one that has authorization to pull the data?
- What if a process AppPackage needs a file reference, but the app needs to pull it, not Weaver? 

This has to be hacked with a CWL type string for now.

There is a need for some flag that tells explicitly `fetch=True|False` under corresponding WPS-REST inputs to force Weaver to do some download operation or not.

WPS has a parameter `asReference=True|False` for outputs (ie `outputTransmission` in more recent versions) that can indicate if the output result has to be dumped as raw data or provided as file reference, but the same parameter for input doesn't seem to exist (at least to my knowledge).

fmigneault:
@tomLandry @dbyrns 
Tagging you for discussion and considerations in future developments.

dbyrns:
We had this discussion in early days of PAVICS and started using string mimetype for url that should be consumed as is (same as what we do with cwl right now). So no they're is no asReference equivalent for inputs. I see two differents issues here:
1. How an app should express its desire to get an url rather that a file access via CWL (Should be enough to make the decision at the ADES level)
2. How the EMS should handle those inputs. The default strategy for the EMS should be to avoid fetching anything unless we know for sure that the source is private and the EMS has the key to it. In which case, via a known private sources list parameter, we could make the decision to fetch the data and offer a temporary public url for downstream applications.

fmigneault:
I've added some details about the *current* operations applied with file reference: 
https://pavics-weaver.readthedocs.io/en/aws-s3/processes.html#file-reference-types

When in situation of ADES, above link describe best the combinations. 
Case (2) is already working as is, in the sense that anything else than a local file is left as plain file reference for ADES called to handle it.

fmigneault:
Since the creation of this issue, Weaver and OGC API evolved a lot. 
Currently, the "standard" approach to define remote contents is using the following definition: 

[...]

PR #184, opened 2020-07-11
AWS S3 support

fmigneault:
- add support of files stored under AWS S3 buckets as process inputs
- add support of process results (output) storage to S3 bucket
- add doc descriptions detailing handling methodology of various file reference types across instance/wps/etc. types


CHANGES.rst
```diff
@@ -8,19 +8,28 @@ Changes
...
+- Add documentation references to `ESGF-CWT Compute API`.
+- Add conditional input file reference fetching (depending on `ADES`/`EMS`, process *type*  from `CWL` ``hints``)
+  to take advantage of *request-options* and all supported scheme formats by `Weaver`, instead of relying on ``PyWPS``
+  and/or ``CWL`` wherever how far downstream the URL reference was reaching.
 
 Fixes:
 ------
-- No change.
+- Adjust some docstrings to better indicate raised errors.
+- Adjust ``weaver.processes.wps_package.WpsPackage`` to use it internal logger when running the process in order to
```
MatProv, line 23: typo: it -> its

config/weaver.ini.example
```diff
@@ -59,6 +59,7 @@ weaver.wps_output = true
 weaver.wps_output_dir = /tmp
 weaver.wps_output_url =
 weaver.wps_output_path = /wpsoutputs
+weaver.wps_output_s3_bucket =
```
MatProv, line 62: where is `weaver. wps_output_s3_region`?

tests/functional/test_wps_package.py
```diff
@@ -1322,3 +1363,136 @@ def test_enum_array_and_multi_format_inputs_from_wps_xml_reference(self):
...
+            "weaver.wps_output_s3_bucket": "wps-output-test-bucket",
+            "weaver.wps_output_s3_region": MOCK_AWS_REGION,  # must match exactly, or mock will not work
+            "weaver.wps_path": "/ows/wps",
+            "weaver.wps_restapi_path": "/",
+        }
+        super(WpsPackageAppWithS3BucketTest, cls).setUpClass()
+
+    @mocked_aws_credentials
+    @mocked_aws_s3
+    def test_execute_with_bucket(self):
```
MatProv, line 1386: `test_execute_process_with_bucket` instead, to make it clearer if, for example, workflow would have particularities that would make the behaviour different

docs/source/configuration.rst
```diff
@@ -68,11 +81,39 @@ Following is a partial list of most predominant settings specific to `Weaver`.
...
     not strictly enforced.
 
+- | ``weaver.wps_output_s3_bucket = <s3-bucket>``
+  | (default: ``None``)
+  |
+  | AWS S3 bucket where to store WPS outputs. Used in conjunction with ``weaver.wps_output_s3_region``.
+  |
+  | When this parameter is defined, any job result generated by a process execution will be stored (uploaded)
+    to that location. If no bucket is specified, the outputs fall back to using the location specified by
+    ``weaver.wps_output_dir``.
```
MatProv, line 91: And what happens if `weaver.wps_output_s3_bucket` is defined but not reachable?
Fallback to `weaver.wps_output_dir` or thows an exception?

fmigneault: It will throw an exception. 
Same as if `weaver.wps_output_dir` points to a directory which the app doesn't have enough permissions to write. 
External setup to weaver provided to it is expected to be configured appropriately, as Weaver cannot do anything about it to fix it anyway.

tests/functional/test_wps_package.py
fmigneault: adjusted to `test_execute_application_package_process_with_bucket` since workflows are also processes

weaver/processes/wps_package.py
```diff
@@ -2089,19 +2124,15 @@ def _handler(self, request, response):
 
             try:
                 self.update_status("Running package...", PACKAGE_PROGRESS_CWL_RUN, STATUS_RUNNING)
-
-                # Inputs starting with file:// will be interpreted as ems local files
-                # If OpenSearch obtain file:// references that must be passed to the ADES use an uri starting
-                # with OPENSEARCH_LOCAL_FILE_SCHEME://
```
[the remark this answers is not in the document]
dbyrns: I saw that you just moved the note into make_location_input which make sense... I wonder why the note was here?

weaver/processes/wps_package.py
```diff
@@ -2137,52 +2197,115 @@ def make_location_input(input_type, input_definition):
...
+        result_fn = os.path.split(result_loc)[-1]
+
+        if s3_bucket:
+            # result_wps = "s3://{}/{}".format(s3_bucket, result_fn)
+            # when 'url' is directly enforced, 'ComplexOutput.json' will use it instead of 'file' from temp workdir
+            # self.response.outputs[output_id].url = result_wps
+
+            # override builder only here so that only results are uploaded to S3, and not XML status also
+            # settings are retrieved from PyWPS server config
+            self.response.outputs[output_id]._storage = S3StorageBuilder().build()  # noqa: W0212
```
dbyrns, line 2269: Wow this add-on to pywps was a gift!

weaver/processes/wps_package.py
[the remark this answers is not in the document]
fmigneault: I guess things got moved around incorrectly at some point. 
I was a bit confused by it also when I modified that part.

weaver/processes/wps_package.py
```diff
@@ -2114,18 +2145,47 @@ def _handler(self, request, response):
...
+        """
+        Figures out if file reference should be fetched immediately for local execution.
+        If anything else than local script/docker, remote ADES/WPS process will fetch it.
+        S3 are handled here to avoid error on remote WPS not supporting it.
+
+        .. seealso::
+            - :ref:`File Reference Types`
+        """
+        if self.is_ems or self.package_type == PROCESS_WORKFLOW:
+            return False
```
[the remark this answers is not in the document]
fmigneault: Indeed there's different behavior happening here whether it is EMS/ADES. 
We could argue that one shouldn't call the process directly via ADES, and even more, remote WPS shouldn't be registered on an ADES in the first place, only on an EMS. 

So the correct behavior would be to forward references directly, which makes sense in my opinion.
We could have  a builtin process whose job is to do s3->https conversion. 
This would handle #183 quite nicely actually.

docs/source/processes.rst
```diff
@@ -353,19 +355,171 @@ Workflow Operations
...
+`EMS` configuration, it will **always** forward the references (regardless of scheme) exactly as provided as input
+of the process execution request, since it assumes it needs to dispatch the execution to another `ADES` remote server,
+and therefore only needs to verify that the file reference is reachable remotely. In this case, it becomes the
+responsibility of this remote instance to handle the reference appropriately. This also avoids potential problems such
+as if `Weaver` as `EMS` doesn't have authorized access to a link that only the target `ADES` would have access to.
+
+When ``CWL`` package defines ``WPS1Requirement`` under ``hints`` for corresponding `WPS-1/2`_ remote processes being
+monitored by `Weaver`, it will skip fetching of ``http(s)``-based references since that would otherwise lead to useless
+double downloads (one on `Weaver` and the other on the `WPS` side). It is the same in case of ``ESGF-CWTRequirement``
+employed for `ESGF-CWT`_ processes. Because these processes do not normally
```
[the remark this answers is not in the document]
fmigneault: Ah! This was supposed to be the part explaining about S3 that gets downloaded to convert as HTTPS when on ADES.
no wonder it was confusing

weaver/processes/wps_package.py
fmigneault: Yes, and I discovered it only by chance while debug stepping through the fetch call! Very lucky.

weaver/processes/wps_package.py
```diff
@@ -2114,18 +2145,47 @@ def _handler(self, request, response):
...
+        app_req = self.get_application_requirement()
+        if app_req["class"] not in [CWL_REQUIREMENT_APP_BUILTIN, CWL_REQUIREMENT_APP_DOCKER]:
+            if input_ref.startswith("s3://"):
+                return True
+            return False
+        return True
```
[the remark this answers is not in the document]
fmigneault: I think I will change this around in following PR with other comment suggestion. 
Builtin app to do S3->HTTP conversion will be the best. If process is called directly with S3, it will forward is like any other HTTP and assume the WPS knows how to handle it, which could actually be the case.

Issue #154, opened 2020-05-24
[Feature] Remove useless/hard-to-maintain de-duplication of cwltool code


fmigneault:
# Context 

Most of the internal work done by `cwltool` is simply an old copy-paste of original code. 
To allow update with most recent `cwltool` patches and features, all dupe code should be totally removed (or refactored as required).  This avoids maintaining it on our side and lets us take advantage of any new development or useful feature from the source.

The only refactors or 'addons' we should provide are our `Job` implementations for remote WPS. All the rest should be handled by `cwltool` procedures natively.

# Changes

For instance: 

-----------
https://github.com/crim-ca/weaver/blob/ede8098736185e87d5c608647cd9a59249739049/weaver/processes/wps_workflow.py#L385 corresponds to : 
`class CommandLineJob(JobBase):`
https://github.com/common-workflow-language/cwltool/blob/7dec97bb8f0bc2d9e9eb710faf41f2e98cc7cdda/cwltool/job.py#L516 
(maybe some minor diffs since)

---------
and 
https://github.com/crim-ca/weaver/blob/ede8098736185e87d5c608647cd9a59249739049/weaver/processes/wps_workflow.py#L182-L189 corresponds to : 
`CommandLineTool.collect_output_ports`
https://github.com/common-workflow-language/cwltool/blob/7dec97bb8f0bc2d9e9eb710faf41f2e98cc7cdda/cwltool/command_line_tool.py#L791

---------
and 
https://github.com/crim-ca/weaver/blob/ede8098736185e87d5c608647cd9a59249739049/weaver/processes/wps_workflow.py#L238-L244 corresponds to: 
`CommandLineTool.collect_outputs`
https://github.com/common-workflow-language/cwltool/blob/7dec97bb8f0bc2d9e9eb710faf41f2e98cc7cdda/cwltool/command_line_tool.py#L877

---------

## **As for below required changes, overrides can be improved/reduced**

---------
At the moment, we override the ``__init__`` and `job` methods to correctly handle the docker requirement and pass down the job to the appropriate remote 'WPS1/3/ESGF' job runner.  
This could be even further simplified. 

If we only override `make_job_runner` to return our WPS dispatch job runners, they would be called by the following: 
```
        j = self.make_job_runner(runtimeContext)(
            builder,
            builder.job,
            self.make_path_mapper,
            self.requirements,
            self.hints,
            jobname,
        )
```
[...]

PR #189, opened 2020-07-17
Fix missing log files

MatProv:
Create an ``stdout.log`` or ``stderr.log`` file in case ``cwltool`` hasn't created it

### DIFF
--- a/weaver/processes/wps_workflow.py
+++ b/weaver/processes/wps_workflow.py
@@ -8,6 +8,7 @@ import shutil
 import tempfile
 from functools import cmp_to_key, partial
 from typing import TYPE_CHECKING, Callable, MutableMapping, Text, cast  # these are actually used in the code
+from pathlib import Path
 
 from cwltool import command_line_tool
 from cwltool.builder import CONTENT_LIMIT, Builder, substitute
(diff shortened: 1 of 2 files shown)


weaver/processes/wps_workflow.py
```diff
@@ -270,6 +271,13 @@ def collect_output(self,
                         try:
                             prefix = fs_access.glob(outdir)
                             key = cmp_to_key(cast(Callable[[Text, Text], int], locale.strcoll))
+
+                            # In case of stdout.log or stderr.log file not created
+                            if glob in ("stdout.log", "stderr.log"):
```
fmigneault, line 276: we should use the literal string but rather the entries in the cwl package as this can have any value

MatProv: You mean using value of `glob`? 
If so then I would suggest to limit ourselves to `stdout.log` and `stderr.log` in order to avoid creating empty file of a glob that should not have an output, in order to raise consequently.

MatProv:
> Can you add a note in #154 referring to these changes?
> I feel this is something that could be easily forgotten during refactor.
> 
> We should also have a unittest that validates the files exist after it is executed.

Yes, I've added it at the bottom of the description

weaver/processes/wps_workflow.py
fmigneault: What I mean is that it shouldn't be hardcoded `"stdout.log"` but `package["stdout"]` or similar to get the real file name placed in the CWL definition since this could be modified at any point.

MatProv: Good point, updated it, let me know if you're alright with the change

fmigneault:
all good 👍

Issue #191, opened 2020-07-20
[BUG] wrong workflow_input_value usage


matprov:
**Describe the bug**
`workflow_input_value` can have multiple types (eg: href, float, int) but `wps3_process.execute()` expects it to be a href.
When it isn't and href, `wps3_process.execute()` fails.
Either when trying `workflow_input_value["location"]` when `workflow_input_value` is a float (which isn't subscriptable), or later on, during the `.startswith()` validation, which fails since type mismatch.

**To Reproduce**
1. Run a workflow, for eg: `pytest tests/functional/test_ems_end2end.py -k "test_workflow_subset_picker"`
2. Notice the TypeError message in the logs (see Screenshots section)

**Expected behavior**
Having data other than `href` parsed as `data`, to avoid type mismatch with `LiteralData`.

**Screenshots**
```
weaver_1        |   File "/opt/local/src/weaver/weaver/processes/wps_workflow.py", line 455, in execute
weaver_1        |     self.results = self.wps_process.execute(self.builder.job, self.outdir, self.expected_outputs)
weaver_1        |   File "/opt/local/src/weaver/weaver/processes/wps3_process.py", line 256, in execute
weaver_1        |     execute_req_input_val: workflow_input_value["location"]})
weaver_1        | TypeError: 'float' object is not subscriptable
```

**Context (please complete the following information):**
 - OS: linux
 - Instance: local EMS
 - Version 1.13.1

Note that although `test_workflow_subset_picker` is not currently working, in the past this test passed successfully on a local EMS.

PR #193, opened 2020-07-22
Fix workflow input value in wp3 process

MatProv:
Related to #191 

- Add ``data`` as a ``workflow_input_value`` data type.
- Fix stdout/stderr log file not present in workflow.


MatProv:
@fmigneault Things might silently break following this change. I need your review on this please.
Type mismatch not present anymore, but can't be fully tested locally with `test_workflow_subset_picker` since `wpsoutputs` in local are not accessible, so getting a 404 in `utils.fetch_file()`

fmigneault:
We should add a test to avoid unexpected errors + detect them in the future.

MatProv:
> We should add a test to avoid unexpected errors + detect them in the future.

I suggest making sure the functional test `test_workflow_subset_picker` gets executed by the CI would do the trick.

fmigneault:
> I suggest making sure the functional test test_workflow_subset_picker gets executed by the CI would do the trick.
This test is currently not ran on Travis.
Running this test makes sure that href as well as data attributes are parsed, since the subset use LiteralData for lat/lon.

I'm a bit confused by what you are running. Looking at the definition of `test_workflow_subset_picker`, the workflow refers to a single process `ColibriFlyingpigeon_SubsetBbox` which is WPS-1 reference. Therefore, the portion of code modified in this PR (WPS-3) should not even be called. It is the class in `weaver/processes/wps1_process.py` that should be executed.

MatProv:
> I'm a bit confused by what you are running. Looking at the definition of `test_workflow_subset_picker`, the workflow refers to a single process `ColibriFlyingpigeon_SubsetBbox` which is WPS-1 reference. Therefore, the portion of code modified in this PR (WPS-3) should not even be called. It is the class in `weaver/processes/wps1_process.py` that should be executed.

Yes, that's a weird behaviour. I actually have no clue why `wp3_process` is getting called instead of `wps1_process`.
Since this PR fixes (well, I guess) the `TypeMismatch`, that means that another misbehaviour might occur beforehand in the process, at the time of selecting which `wpsX_process` version to use.

fmigneault:
> Yes, that's a weird behaviour. I actually have no clue why `wp3_process` is getting called instead of `wps1_process`.
> Since this PR fixes (well, I guess) the `TypeMismatch`, that means that another misbehaviour might occur beforehand in the process, at the time of selecting which `wpsX_process` version to use.

If this is an actual issue, it could be investigated using another PR/test case. 
Regardless of that, I would like to have more "targeted" tests that verify the feature/fix added here. 
I don't like the idea of considering the fix as tested through a huge workflow which imply a lot of other functionalities (and other potential cause of error).

Basically, we need clear and simple test that runs `Wps3Process.execute` (as if CWL was calling it to run a job), and verify that inputs of different types are parsed as expected. Using a mock of `WpsProcessInterface.make_request`, we can simulate the responses of each step without actually calling a real server.

MatProv:
> If this is an actual issue, it could be investigated using another PR/test case.
> Regardless of that, I would like to have more "targeted" tests that verify the feature/fix added here.
> I don't like the idea of considering the fix as tested through a huge workflow which imply a lot of other functionalities (and other potential cause of error).
> 
> Basically, we need clear and simple test that runs `Wps3Process.execute` (as if CWL was calling it to run a job), and verify that inputs of different types are parsed as expected. Using a mock of `WpsProcessInterface.make_request`, we can simulate the responses of each step without actually calling a real server.

#195 will get rid of that misbehaviour related to WPS1-WPS3.
And yes, a targeted test will be added in order to validate input parsing.

fmigneault (approved):
Added missing test + some small patch for new change by AWS S3

PR #206, opened 2021-01-26
support YAML for weaver.data_sources

fmigneault:


fmigneault:
Merging as target python version tests pass. 
Only older / experimental python versions failing

fmigneault:
Merging as target python version tests pass.
Only older / experimental python versions failing

Issue #198, opened 2020-08-18
[BUG] Failing CWT install of ESGF-compute API


fmigneault:
**Describe the bug**

The requirement specified here: 
https://github.com/crim-ca/weaver/blame/d156e01dc9bffa107aa2d3c2fcb3f720d7963c97/setup.py#L53

Constantly results into following error when installing weaver from fresh (or update): 

```
[...]
  File "/home/francis/DEV/miniconda/envs/weaver-py3/lib/python3.7/site-packages/pkg_resources/__init__.py", line 899, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/home/francis/DEV/miniconda/envs/weaver-py3/lib/python3.7/site-packages/pkg_resources/__init__.py", line 785, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'cwt' distribution was not found and is required by weaver
Exception ignored in: <module 'threading' from '/home/francis/DEV/miniconda/envs/weaver-py3/lib/python3.7/threading.py'> 
```

The requirement doesn't seem to like very much the different `cwt` vs `esgf-compute-api` package installed.

Furthermore, looking at https://github.com/ESGF/esgf-compute-api/tree/v2.1.0, the `cwt.process` also contain some `print status` operation which is Python-2. 
This is only fixed in v2.2.0, but there are many (potentially breaking) changes also. 
Maybe we should try to request a v2.1.1 with only this statement removed.

In any way, this requirement is extremely breaking weaver install, and must be resolved quickly.

MatProv:
@fmigneault Did it happened during local installation or on brand new instances?
I wonder why during installation of this dependency on Travis we don't have this kind of issue, and how it differs from a fresh install.

MatProv:
Also, I would say that upgrading 2.1.0 to 2.2.0 would not be breaking things up in our case since this dependency has been added really recently in weaver. 

Version 2.1.0 was chosen instead of 2.2.0 only because 2.2.0 wasn't out when the original `esgf-cwt-process` branch has been created, not because 2.2.0 was unsuitable for our needs. 

I think that in our case, going from 2.1.0 to (potentially) 2.1.1, then 2.2.0 would be the same thing as going from 2.1.0 to 2.2.0. Running existing tests to make sure 2.2.0 doesn't break anything in weaver would be the only assertion we can make, since 2.1.0 hadn't been chosen for particular reason, other than "latest available tag".

[...]

PR #207, opened 2021-01-27
Resolve requirements

fmigneault:
- temp fix for https://github.com/geopython/pywps/issues/568 (mismatch python 2/3 deps)
  resolved via targeting https://github.com/geopython/pywps/pull/569
- patch esgf-cwt api that still referenced a version that uses `print` without parenthesis (python 2 style)
- handle pip 21.0 that will raise eventually due to `--use-feature=2020-resolver`
  `make` will add it on the fly as needed

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,7 +12,8 @@ Changes:
 
 Fixes:
 ------
-- No change.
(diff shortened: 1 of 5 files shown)


fmigneault:
Fixes #198

Issue #126, opened 2020-05-07
[Feature] Handle WPS-1.x/2.x execute request as WPS-3 'job'


fmigneault:
Route `/ows/wps` (default path) offers the WPS-1.x/2.x XML interface (maybe also its 2.0 JSON interface with #125). 

Since this route is served as a standard PyWPS service, it is possible to call an `Execute` request (either with GET/POST variants), but this will run the process directly on an API 'manager' thread instead of some dedicated celery worker thread/docker image. This both consumes the few resources for handling REST-API requests and possibly limit how fast/many job request the actual API can process, as each 'manager' thread becomes synchronously blocked/used until the corresponding process completes (could be a very long workflow). 

Even worst, because the WPS process is executed directly from the PyWPS service, no actual `job` gets registered in `Weaver` and it is therefore impossible to obtain any feedback from the WPS 3.x routes. The only access to process status is via the base PyWPS status URL, which will consume yet another API-manager thread. Finally, when the process completes, no log/exceptions/results will be available, except the bare WPS Status route.

Therefore, WPS-1/2 Execute requests should be redirected to the corresponding `/processes/{id}/jobs` call, so that the job gets created and processed using the normal celery worker that will pick it up eventually. Feature #21 must be implemented at the same time as the redirect will have to work the other way around of current implementation (ie: WPS-1/2 Execute => WPS-REST 'celery' execute). 
Without simultaneous implementation of the features, we will generate a circular loop of execute calls between implementations.

Given that redirections between WPS-1/2 and WPS-3 calls would be implemented here, features in #125 could be done at the same time as it implies similar requirements.

Relates to double-layer monitoring #21.

Issue #21, opened 2019-03-15
Remove double process execution monitoring layers


fmigneault:
At the moment, processes are executed via monitoring the WPS-1 endpoint provided by `pywps` with an `owslib.WebProcessingService`. The `pywps` also monitors the sub-job execution, creating unnecessary layers of job monitoring, harder debugging, and complicated logging.

A single `celery` job monitor should be done, directly monitoring the real job execution of the process.
Also, the need to pass by the `pywps` endpoint should be removed for job execution.

The WPS-1 endpoint should be preserved for the sole purpose of providing back-compatibility with WPS-1 by producing a sub-request to the WPS-3 REST-API for following cases.
- GetCapabilities => `GET /processes`
- DescribeProcess => `GET /processes/{id}`
- Job Execute => `POST /processes/{id}/jobs`
- Job GetStatus (n/a) use `GET /processes/{id}/jobs/{id}` instead

PR #209, opened 2021-02-02
WPS worker jobs

fmigneault:
rebased on #207 and #208 to align worker image to run docker packages

- partially address #21 and #126 
  Weaver `worker` is now the one running the job (tested with a dummy docker echo) - not the API that was originally receiving the final execute as it hosts the PyWPS endpoint.

- more tests needed for remote WPS, but they should technically run the same way as before, except under an additional loop of celery async task

- still needs some work to complete the redirection of `GET` / `POST` WPS Execute requests directly on PyWPS endpoint, such that they also go through the same celery worker job dispatching. 
  - must add parsing of XML => JSON to forward request. 
  - maybe `POST /processes/{id}/jobs` could directly handle this format based on `Content-Type` header
  - check if possible to reuse some of the util methods in `wps_package` (eg: `_wps2json` and other transforms)


dbyrns (approved):
LGTM

MatProv (approved):

PR #219, opened 2021-02-24
WPS load config

fmigneault:
- Ensure that configuration file definitions specified in ``processes`` and ``providers`` will override older database
  definitions respectively matched by ``id`` and ``name`` when starting `Weaver` if other parameters were modified.


dbyrns (dismissed):
LGTM

PR #221, opened 2021-02-26
Job io extra 2

fmigneault:
## Context

Recent version `2.x` fixes brought back into branch [job-io-extras](https://github.com/crim-ca/weaver/compare/job-io-extras) and completed to make all tests pass. 

## Changes

- many conformance fixes related to changes in [opengeospatial/ogcapi-processes](https://github.com/opengeospatial/ogcapi-processes)
  - most notable: 
    - `jobs/{id}/result` is deprecated but redirects to `jobs/{id}/outputs` which as similar format (but with extra `links` references)
    - `jobs/{id}/results` follows the official output format of OGC-API schemas
    - conformance path updated with links to schema-class of OGC-API (complete tests todo in later PR #231)
- more explicit OpenAPI (v3) schema definitions with custom `cornice_swagger` & `colander` generators 
  - swagger v2 was used before, with no possible definition of all `oneOf`, `allOf`, etc. keywords
  - explore: https://pavics-weaver.readthedocs.io/en/job-io-extra-2/api.html
    vs "old" one: https://pavics-weaver.readthedocs.io/en/2.2.0/api.html
    notable differences in Deploy body for example.
- *enforced validation* of input/output contents against *same* schema deserialization from OpenAPI 
  - ensures that structures are exactly as documented (what gets deployed/executed is what is documented)
  - expected value formats, limits by min/max, etc. are all pre-validated inplace by schemas
  - provides detailed indication of why a structure is invalid in case of error 
    (eg `field x is invalid because ... expected ... `) 
    with deeply nested parsing (as far as provided by schemas)  
- other features that were explored some long time ago but still not integrated: 
  - GPU-enabled docker app and 
  - Windows execution 
  - Caching of some static endpoints

### Detailed schema definitions (before vs after) : 
image

### Invalid schema validation error message

Specific error raised details which field, index, value, etc. cause the error and everything that was evaluated against valid schemas.
Error is both logged and reported in corresponding request response. 
In this case, invalid value `random` was placed as `type` for the one of the CWL package inputs.

image

### Documented deprecation / changes

image


dbyrns (approved):
Great conformance job! Just be careful with feature creep (Windows support, requests caching)

PR #225, opened 2021-03-02
fix job log reporting with gradual progress completion

### LINKED ISSUE #90: A job's 'percentCompleted' is not the same percent as in the status message
During a WPS1 execution, we can see the percentage at 2 places: in the `percentCompleted` field, and in the `message` field. The value in the message field seems to be the good one.

At one point, the percent shown in the message was 90% (right at the end of the processing, so `REMOTE_JOB_PROGRESS_FETCH_OUT` was the right value), and in the `percentCompleted` field, it was 30%.

After a quick look at the code, I think that the `weaver.process.utils.map_progress` function is called twice (in 2 different places) before being outputted in `percentCompleted`, but once for the value in the status message.

So this one is ok:
https://github.com/crim-ca/weaver/blob/339e57b7a03a82e840c6068ce5919b131afef48f/weaver/processes/wps1_process.py#L132

But this one should treat the progress value as the final percentage I believe (or at least in this case):
https://github.com/crim-ca/weaver/blob/339e57b7a03a82e840c6068ce5919b131afef48f/weaver/processes/wps_package.py#L1709

fmigneault:
fixes #90

Job execution logs side-by-side comparison of before/after changes. 
Instead of reporting same (final) percent all the time and status `failed`/`succeeded` accordingly to final result, reports `running` (which was the status of the job at the time they were logged) and make percent proportionally increase. 

Sadly, reported progress/status is still limited to how verbose the package is *during* its execution - and collected after the fact to apply to the job log, but they are more consistent with what happened otherwise. 

image

### DIFF
--- a/weaver/processes/execution.py
+++ b/weaver/processes/execution.py
@@ -209,7 +209,7 @@ def execute_process(self, job_id, url, headers=None):
                         wps_package.retrieve_package_job_log(execution, job, progress_min, progress_max)
                         job.status = map_status(STATUS_SUCCEEDED)
                         job.status_message = "Job succeeded{}.".format(msg_progress)
-                        job.progress = JOB_PROGRESS_EXECUTE_MONITOR_DONE
+                        job.progress = progress_max
                         job.save_log(logger=task_logger)
                         job_results = [ows2json_output_data(output, process, settings)
                                        for output in execution.processOutputs]
@@ -218,7 +218,7 @@ def execute_process(self, job_id, url, headers=None):
                         task_logger.debug("Job failed.")
                         wps_package.retrieve_package_job_log(execution, job, progress_min, progress_max)
                         job.status_message = "Job failed{}.".format(msg_progress)
-                        job.progress = JOB_PROGRESS_EXECUTE_MONITOR_DONE
+                        job.progress = progress_max
                         job.save_log(errors=execution.errors, logger=task_logger)
                     task_logger.debug("Mapping Job references with generated WPS locations.")
                     map_locations(job, settings)
--- a/weaver/processes/execution.py
+++ b/weaver/processes/execution.py
@@ -59,15 +59,16 @@ if TYPE_CHECKING:
 # job process execution progress
 JOB_PROGRESS_SETUP = 1
 JOB_PROGRESS_DESCRIBE = 2
-JOB_PROGRESS_GET_INPUTS = 4
-JOB_PROGRESS_GET_OUTPUTS = 6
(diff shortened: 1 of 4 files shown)


weaver/processes/execution.py
```diff
@@ -201,20 +204,21 @@ def execute_process(self, job_id, url, headers=None):
...
                 if execution.isComplete():
                     job.mark_finished()
-                    job.progress = JOB_PROGRESS_EXECUTE_MONITOR_END
                     msg_progress = " (status: {})".format(job_msg) if job_msg else ""
                     if execution.isSucceded():
+                        wps_package.retrieve_package_job_log(execution, job, progress_min, progress_max)
                         job.status = map_status(STATUS_SUCCEEDED)
                         job.status_message = "Job succeeded{}.".format(msg_progress)
-                        wps_package.retrieve_package_job_log(execution, job)
+                        job.progress = JOB_PROGRESS_EXECUTE_MONITOR_DONE
```
MatProv, line 212: Why not use `progress_max`? Same comment for below usage of `JOB_PROGRESS_EXECUTE_MONITOR_DONE `.

weaver/processes/wps_package.py
```diff
@@ -164,8 +164,13 @@ def retrieve_package_job_log(execution, job):
         # if the process is a weaver package this status xml should be available in the process output dir
         log_path = get_status_location_log_path(execution.statusLocation, out_dir=out_dir)
         with open(log_path, "r") as log_file:
-            for line in log_file:
-                job.save_log(message=line.rstrip("\n"))
+            log_lines = log_file.readlines()
+        if not log_lines:
+            return
```
MatProv, line 169: Simply return or throw exception?

weaver/processes/wps_package.py
```diff
@@ -164,8 +164,13 @@ def retrieve_package_job_log(execution, job):
...
+        total = float(len(log_lines))
+        for i, line in enumerate(log_lines):
+            progress = map_progress(i / total * 100, progress_min, progress_max)
```
MatProv, line 172: What if `len(log_lines) == 0`?

MatProv (approved):

weaver/processes/wps_package.py
fmigneault: Simply return. If the process was not verbose, we won't break its execution for it. Logs will just be minimal.

dbyrns (commented):
I wanted to let you know that I still validate and saw nothing wrong.

PR #226, opened 2021-03-02
add get_wps_client function with request options

fmigneault:
Employs a common generator of `WebProcessingService` client to apply requests options in a more consistent way.


MatProv (approved):

dbyrns (commented):
Nice reuse! That must help a lot.

PR #232, opened 2021-03-17
wps-client cached

fmigneault:
wps-client creation caching control according to beaker config and request headers

Avoids re-querying the full process listing and client object such that multiple calls of `/providers/x/processes/y`, `/providers/x/processes/z`, etc. will only do an initial `getcapabilites`, and will reuse the object for subsequent `describeprocess` operation until the cache expires or is enforced with no-cache.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,7 +10,8 @@ Changes
 
 Changes:
 --------
-- No change.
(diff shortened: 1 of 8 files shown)


dbyrns (approved):
LGTM

PR #241, opened 2021-04-01
Schema xml+examples

fmigneault:
## changes

- apply multiple updates to OpenAPI schema definitions to provide examples
- align definitions of jobs results, outputs, etc. with recent route response formats 
- provide XML definitions of that endpoint using added support of XML attributes
- improve resolution of schema as `$ref` (every single OpenAPI definition can be more easily mapped with high reliability to original Python class) 

## test

following doc can be used to look as result from this PR 
https://pavics-weaver.readthedocs.io/en/schema-xmlexamples/api.html

## examples

WPS endpoint parameters: 

image

WPS XML body response

image

Integrated XML examples 
image

And JSON examples
image

Snipets of definitions for JSON, XML and CWL definitions: 
image


dbyrns (approved):
LGTM. Do we have a spec version or something that we could refer so that we know on what we are compliant?

fmigneault:
> 
> Do we have a spec version or something that we could refer so that we know on what we are compliant?

For the moment, *OGC-API Processes* is still working in terms of drafts, so nothing is final.
Next one planed is [draft6](https://github.com/opengeospatial/ogcapi-processes/issues?q=is%3Aissue+is%3Aopen+label%3A1.0-draft.6) and should be coming out in the following weeks or so. 

There are also some considerations around a test-suite of API implementations.
https://github.com/opengeospatial/ogcapi-processes/issues/157
That would be the only reliable way to validate minimal requirements are fulfilled on OGC-API side. 
As for WPS-1/2 compliance, there is not many useful resources (but not that much a problem since we are phasing out of them?).

dbyrns:
I know that nothing is final, but I was wondering on which document the latest revision was based upon. 
Indeed external test-suite for API validation would be great.
And for WPS-1/2, I'm not really worried as we already used owslib as client.

PR #258, opened 2021-06-03
fix invalid Not keyword resolution 

fmigneault:
fix invalid Not keyword resolution for `LiteralInputType` which resulted into too verbose process description inputs

The missing deserialization of the resulting mapping following OpenAPI `not` keyword validation resulted into all fields under a literal input type to bubble up in the process description. For example, fields like `data_format`, `mode`, `mimetype`, `workdir`, etc. that come from the pywps/owslib/cwl object json conversion would be returned.

### DIFF
--- a/tests/functional/test_wps_package.py
+++ b/tests/functional/test_wps_package.py
@@ -48,8 +48,15 @@ IANA_TAR = IANA_NAMESPACE + ":" + CONTENT_TYPE_APP_TAR  # noqa # pylint: disable
 IANA_ZIP = IANA_NAMESPACE + ":" + CONTENT_TYPE_APP_ZIP  # noqa # pylint: disable=unused-variable
 
 KNOWN_PROCESS_DESCRIPTION_FIELDS = {
-    "id", "title", "abstract", "inputs", "outputs", "executeEndpoint", "keywords", "metadata", "visibility"
+    "id", "title", "abstract", "keywords", "metadata", "inputs", "outputs", "executeEndpoint", "visibility"
 }
+# intersection of fields in InputType and specific sub-schema LiteralInputType
+KNOWN_PROCESS_DESCRIPTION_INPUT_DATA_FIELDS = {
+    "id", "title", "abstract", "keywords", "metadata", "links", "literalDataDomains", "additionalParameters",
+    "minOccurs", "maxOccurs"
+}
+# corresponding schemas of input, but min/max occurs not expected
+KNOWN_PROCESS_DESCRIPTION_OUTPUT_DATA_FIELDS = KNOWN_PROCESS_DESCRIPTION_INPUT_DATA_FIELDS - {"minOccurs", "maxOccurs"}
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


tests/functional/test_wps_package.py
```diff
@@ -48,8 +48,15 @@
...
 
 KNOWN_PROCESS_DESCRIPTION_FIELDS = {
-    "id", "title", "abstract", "inputs", "outputs", "executeEndpoint", "keywords", "metadata", "visibility"
+    "id", "title", "abstract", "keywords", "metadata", "inputs", "outputs", "executeEndpoint", "visibility"
 }
+# intersection of fields in InputType and specific sub-schema LiteralInputType
+KNOWN_PROCESS_DESCRIPTION_INPUT_DATA_FIELDS = {
+    "id", "title", "abstract", "keywords", "metadata", "links", "literalDataDomains", "additionalParameters",
+    "minOccurs", "maxOccurs"
+}
```
trapsidanadir, line 57: How about using `set.intersection()`, it might be better to avoid repetition

tests/functional/test_wps_package.py
```diff
@@ -134,6 +141,11 @@ def test_literal_io_from_package(self):
         assert "maxOccurs" not in desc["process"]["outputs"][0]
         assert "format" not in desc["process"]["outputs"][0]
         assert len(set(desc["process"].keys()) - KNOWN_PROCESS_DESCRIPTION_FIELDS) == 0
+        # make sure that deserialization of literal fields did not produce over-verbose metadata
+        for p_input in desc["process"]["inputs"]:
+            assert len(set(p_input) - KNOWN_PROCESS_DESCRIPTION_INPUT_DATA_FIELDS) == 0
```
trapsidanadir, line 146: Using directly `assert not set1 - set2` might be a good way to verify too

tests/functional/test_wps_package.py
```diff
@@ -134,6 +141,11 @@ def test_literal_io_from_package(self):
...
+        for p_output in desc["process"]["outputs"]:
+            assert len(set(p_output) - KNOWN_PROCESS_DESCRIPTION_OUTPUT_DATA_FIELDS) == 0
```
trapsidanadir, line 148: Same as last comment

tests/functional/test_wps_package.py
fmigneault: Can you elaborate on this? I'm not sure which repetition you refer to?

trapsidanadir: Some of the values are already defined at line 51 in the `KNOWN_PROCESS_DESCRIPTION_FIELDS ` set, since you are doing an intersection, of fields, as described in line 53, the suggestion is to use the `set.intersection()` method with the set above and a set of `LiteralInputType`.
If you update one of the fields of a set, the intersection will be updated automatically 😄

fmigneault: Ok. I see what you mean. The values in the process description are not the ones being intersected with though. 
They just match the same names by chance, but are in completely different location in the body.

tests/functional/test_wps_package.py
fmigneault: not critical

PR #260, opened 2021-06-08
fix missing jobControlOptions in pre-deployed processes

fmigneault:
support pre-deployed processes that did not provide jobControlOptions by providing the default value `["async-execute"]`


fmigneault:
FYI @trapsidanadir 
Update for backward support.

trapsidanadir (approved):
Looks great to me !!

PR #262, opened 2021-06-10
Added array support for job input (fixes #233)

### LINKED ISSUE #233: [Feature] Input representation against maxOccurs definition
When `maxOccurs=1`, input should *always* be provided as object.
When `maxOccurs>1`, input should *always* be provided as list of objects.

Schema must be validated/updated and tests created to guarantee the formats are correctly defined and handled when converting between older WPS-1/2 to WPS-REST.

relates to https://github.com/opengeospatial/ogcapi-processes/issues/129

trapsidanadir:


CHANGES.rst
```diff
@@ -13,6 +13,9 @@ Changes:
 - Add reference link to ReadTheDocs URL of `Weaver` in API landing page.
 - Add references to `OGC-API Processes` requirements and recommendations for eventual conformance listing
   (relates to `#231 <https://github.com/crim-ca/weaver/issues/231>`_).
+- Updated the format of the URL for the job results.
+- Add support for array type as job inputs
+  (relates to `#233 <https://github.com/crim-ca/weaver/issues/233>`_).
```
fmigneault, line 18: fixes instead of relates

weaver/wps_restapi/colander_extras.py
```diff
@@ -176,6 +176,12 @@ def __init__(self, *args, v_prefix=False, rc_suffix=True, **kwargs):
 
 
 class ExtendedBoolean(colander.Boolean):
+
+    def __init__(self, *args, true_choices=None, **kwargs):
+        if true_choices is None:
+            true_choices = ("true", "1")
+        super(ExtendedBoolean, self).__init__(true_choices=true_choices, *args, **kwargs)
```
fmigneault, line 183: Add a changelog entry related to that. 
It is important to indicate that values will not be automatically converted to `bool` anymore. 
They will require explicit `false|true`.

In order to align with OpenAPI definitions
> type: boolean represents two values: true and false. Note that truthy and falsy values such as "true", "", 0 or null are not considered boolean values. 

Let's remove the `"1"` and also override the `false_choices` to be only `"false"`. 

I would like also a comment above those changes that indicate the reason they are defined as `"true"` and `"false"` is because colander converters the value to string lowercase to compare with other thruty/falsy values it should accept. 
The comment should warn not to add other values like `"1"`, etc. to avoid conflict with Integer type for schemas that support both variants.

weaver/wps_restapi/swagger_definitions.py
```diff
@@ -717,6 +717,40 @@ class AnyLiteralType(OneOfKeywordSchema):
...
+    value = ExtendedSchemaNode(Float())
+
+
+class ArrayLiteralType(OneOfKeywordSchema):
+    """
+    .. seealso::
+        - :class:`SequenceNumberType`
+        - :class:`SequenceBooleanType`
+        - :class:`SequenceStringType`
+    """
```
fmigneault, line 738: Can be removed since they are referenced right below in the `_one_of`.

weaver/wps_restapi/swagger_definitions.py
```diff
@@ -2026,13 +2060,16 @@ class Reference(ExtendedMappingSchema):
 class AnyType(OneOfKeywordSchema):
     """Permissive variants that we attempt to parse automatically."""
     _one_of = [
+
```
fmigneault, line 2063: remove extra line

weaver/wps_restapi/swagger_definitions.py
```diff
@@ -8,7 +8,7 @@
 from typing import TYPE_CHECKING
 
 import yaml
-from colander import DateTime, Email, OneOf, Range, Regex, drop, required
+from colander import DateTime, Email, OneOf, Range, Regex, SequenceSchema, drop, required
```
fmigneault, line 11: Don't use the original sequence schema, but the `ExtendedSequenceSchema`

tests/functional/test_wps_package.py
```diff
@@ -913,6 +916,120 @@ def test_valid_io_min_max_occurs_as_str_or_int(self):
...
+            proc_url = "/processes/{}/jobs".format(self._testMethodName)
+            resp = mocked_sub_requests(self.app, "post_json", proc_url, timeout=5,
+                                       data=exec_body, headers=self.json_headers, only_local=True)
+            assert resp.status_code in [200, 201], "Failed with: [{}]\nReason:\n{}".format(resp.status_code, resp.json)
+            status_url = resp.json.get("location")
+
+        tmpdir = tempfile.gettempdir()
+        job_id = status_url.split("/")[-1]
+        tmpfile = "{}/{}/tmp.txt".format(tmpdir, job_id)
+
```
fmigneault, line 1018: Add `"weaver.wps_output_dir": "/tmp",  # nosec: B108 # don't care hardcoded for test` to the settings in `setUpClass` (like other places do), and reuse that value here to figure out the base tmp dir.

tests/functional/test_wps_package.py
```diff
@@ -913,6 +916,120 @@ def test_valid_io_min_max_occurs_as_str_or_int(self):
...
+        try:
+            processed_values = json.load(open(tmpfile, "r"))
+        except FileNotFoundError:
+            raise FileNotFoundError
```
fmigneault, line 1022: call `self.fail` message indicating that output file `[<path>]` was not found where it was expected to resume test

fmigneault (changes requested):

CHANGES.rst
```diff
@@ -36,6 +36,8 @@ Changes:
 - Add reference link to ReadTheDocs URL of `Weaver` in API landing page.
 - Add references to `OGC-API Processes` requirements and recommendations for eventual conformance listing
   (relates to `#231 <https://github.com/crim-ca/weaver/issues/231>`_).
+- In order to align with OpenAPI `boolean` type definitions, non explicit `boolean` values will not be automatically 
+  converted to `bool` anymore. They will require explicit `false|true`.
```
fmigneault, line 40: reword with
> Removed implicit ``boolean`` conversion from boolean-like values (``0``, ``1``, ``"true"``, etc.) and enforce explicit `true|false` values wherever applicable to align schema parsing with OpenAPI ``boolean`` type definitions.

Make sure to use double `` symbol for RST parsing to use code-style.

CHANGES.rst
```diff
@@ -51,6 +53,10 @@ Fixes:
 - Fix default execution mode specification in process job control options
   (fixes `#182 <https://github.com/opengeospatial/ogcapi-processes/pull/182>`_).
 - Fix old OGC-API WPS REST bindings link in landing page for the more recent `OGC-API Processes` specification.
+- Updated the format of the URL for the job results.When the prefix `/` was not present, 
+  URL was incorrectly handled by not prepended the required base URL location.
```
fmigneault, line 57: prepending

CHANGES.rst
```diff
@@ -51,6 +53,10 @@ Fixes:
...
+- Add support for array type as job inputs
+  (relates to `#233 <https://github.com/crim-ca/weaver/issues/233>`_).
```
fmigneault, line 59: move to latest changes section

tests/functional/test_wps_package.py
```diff
@@ -925,6 +930,174 @@ def test_valid_io_min_max_occurs_as_str_or_int(self):
...
+                "test_string_array": {"type": {"type": "array", "items": "string"}},
+                "test_reference_array": {"type": {"type": "array", "items": "File"}},
+                "test_int_value": "int",
+                "test_float_value": "float",
+                "test_string_value": "string",
+                "test_reference_http_value": "File",
+                "test_reference_file_value": "File",
+                "test_reference_s3_value": "File"
+            },
+            "requirements": {
```
fmigneault, line 956: Add the method we confirmed works using docker-python image.

tests/utils.py
```diff
@@ -549,3 +549,25 @@ def mocked_aws_s3_bucket_test_file(bucket_name, file_name, file_content="Test fi
...
+
+def mocked_http_test_file(dirname, file_url, file_content="This is a generated file for http test"):
+    # type: (str,str,str) -> str
+    """
+    Generates a test file reference from dummy data
+    """
+    path = dirname+file_url.split("localhost")[-1]
+    with open(path, "w") as tmp_file:
+        tmp_file.write(file_content)
+    return path
```
fmigneault, line 562: This doesn't feel right. HTTP is not actually involved anywhere in this. 

The `http://` URL you provide as input in the test gets replaced right here, so the job execution that receives the returned path is a local file path, which is not different that the other case that gives directly `file://`.

tests/utils.py
```diff
@@ -549,3 +549,25 @@ def mocked_aws_s3_bucket_test_file(bucket_name, file_name, file_content="Test fi
...
+
+def mocked_file_test(dirname, file_path, file_content="This is a generated file for file test"):
+    # type: (str,str,str) -> str
+    """
+    Generates a test file reference from dummy data
+    """
+    path = dirname+"/"+file_path
+    with open(path, "w") as tmp_file:
+        tmp_file.write(file_content)
+    return "file://{}".format(path)
```
fmigneault, line 573: I don't think there is sufficient operations involved here to be worth having a dedicated function for it. 
The test itself could just dump the file directly where needed. 

Also, avoid using "mocked" wording here, nothing is being mocked. This makes the test definition more confusing.

weaver/processes/execution.py
```diff
@@ -118,7 +118,11 @@ def execute_process(self, job_id, url, headers=None):
                 input_values = process_value if isinstance(process_value, list) else [process_value]
 
                 # we need to support file:// scheme but PyWPS doesn't like them so remove the scheme file://
-                input_values = [val[7:] if str(val).startswith("file://") else val for val in input_values]
+                input_values = [
+                    (get_any_value(val)[7:] if str(get_any_value(val)).startswith("file://") else get_any_value(val))
+                    if isinstance(val, dict) else (val[7:] if str(val).startswith("file://") else val)
+                    for val in input_values
+                ]
```
fmigneault, line 125: please reformat as such to make it easier to read individual if cases
```
input_values = [
        # when value is an array of dict that each contain a file reference
	(get_any_value(val)[7:] if str(get_any_value(val)).startswith("file://") else get_any_value(val)) 
        if isinstance(val, dict) else 
        # when value is directly a single dict with file reference
        (val[7:] if str(val).startswith("file://") else val)
	for val in input_values
]
```

weaver/wps_restapi/colander_extras.py
```diff
@@ -176,6 +176,19 @@ def __init__(self, *args, v_prefix=False, rc_suffix=True, **kwargs):
 
 
 class ExtendedBoolean(colander.Boolean):
+
+    def __init__(self, *args, true_choices=None, fase_choices=None, **kwargs):
+        """
+        The arguments true_choices and false_choices are defined as `"true"` and `"false"` since
```
fmigneault, line 182: use ``` :paramref:`true_choices` ``` and double `` for true/false

weaver/wps_restapi/colander_extras.py
```diff
@@ -176,6 +176,19 @@ def __init__(self, *args, v_prefix=False, rc_suffix=True, **kwargs):
...
+        colander converts the value to string lowercase to compare with other thruty/falsy values it should accept.
+        Do NOT add other values like "1" to avoid conflict with Integer type for schemas that support both variants.
```
fmigneault, line 184: use ``` :paramref:`true_choices` ``` 
use ``` :mod:`colander` ``` 
use double `` for true/false, "1" and Integer

fmigneault (changes requested):
Some more work is needed regarding the test. 
Otherwise, some formatting issues here and there to adjust.

CHANGES.rst
```diff
@@ -54,7 +55,7 @@ Fixes:
   (fixes `#182 <https://github.com/opengeospatial/ogcapi-processes/pull/182>`_).
 - Fix old OGC-API WPS REST bindings link in landing page for the more recent `OGC-API Processes` specification.
 - Updated the format of the URL for the job results.When the prefix `/` was not present, 
-  URL was incorrectly handled by not prepended the required base URL location.
+  URL was incorrectly handled by not prepending the required base URL location.
```
fmigneault: indications from last review are still not fixed for this part.

CHANGES.rst
```diff
@@ -51,6 +54,10 @@ Fixes:
 - Fix default execution mode specification in process job control options
   (fixes `#182 <https://github.com/opengeospatial/ogcapi-processes/pull/182>`_).
 - Fix old OGC-API WPS REST bindings link in landing page for the more recent `OGC-API Processes` specification.
+- Updated the format of the URL for the job results.When the prefix `/` was not present,
```
fmigneault, line 57: space after results. double quote on ``/``

CHANGES.rst
```diff
@@ -33,9 +34,11 @@ Fixes:
 
 Changes:
 --------
-- Add reference link to ReadTheDocs URL of `Weaver` in API landing page.
-- Add references to `OGC-API Processes` requirements and recommendations for eventual conformance listing
+- Add reference link to ReadTheDocs URL of ``Weaver`` in API landing page.
+- Add references to ``OGC-API Processes`` requirements and recommendations for eventual conformance listing
```
fmigneault, line 38: these two items should remain single quote for italic terms because they are not code

tests/utils.py
```diff
@@ -549,3 +550,38 @@ def mocked_aws_s3_bucket_test_file(bucket_name, file_name, file_content="Test fi
         tmp_file.flush()
         s3.upload_file(Bucket=bucket_name, Filename=tmp_file.name, Key=file_name)
     return "s3://{}/{}".format(bucket_name, file_name)
+
+
+def mocked_http_file(test_func):
+    # type: (Callable[[...], Any]) -> Callable
+    """
+    Mocked http file for http reference test
```
fmigneault, line 558: Please provide a better description. It should indicate what is the purpose/outcome of using this mock.

Refer to `mocked_aws_s3` description. 
There should be a note regarding the specific value defined by `MOCK_HTTP_REF`.

tests/utils.py
```diff
@@ -549,3 +550,38 @@ def mocked_aws_s3_bucket_test_file(bucket_name, file_name, file_content="Test fi
...
+    def wrapped(*args, **kwargs):
+        with mock.patch("weaver.processes.wps_package.fetch_file", side_effect=mocked_file_request):
+            return test_func(*args, **kwargs)
+    return wrapped
+
+
+def mocked_reference_test_file(file_name, href_type, file_content="This is a generated file for href test"):
+    # type: (str,str,str) -> str
+    """
+    Generates a test file reference from dummy data for http and file href types
```
fmigneault, line 578: missing period

fmigneault (changes requested):
Minor formatting issues left. Once fixed, should be good to merge.

tests/utils.py
trapsidanadir: Tried to explain better, can you take a look at the updated comment.

tests/utils.py
```diff
@@ -549,3 +550,40 @@ def mocked_aws_s3_bucket_test_file(bucket_name, file_name, file_content="Test fi
...
         s3.upload_file(Bucket=bucket_name, Filename=tmp_file.name, Key=file_name)
     return "s3://{}/{}".format(bucket_name, file_name)
+
+
+def mocked_http_file(test_func):
+    # type: (Callable[[...], Any]) -> Callable
+    """
+    Creates a mock of the function `fetch_file`, to fetch a generated file locally, for test purposes only.
+    For instance, calling this function with :func:`mocked_http_file` decorator
+    will effectively employ the mocked `fetch_file` and return a generated local file.
```
fmigneault, line 560: please use ```:func:`fetch_file` ```

fmigneault (changes requested):
New *Changes* items are still under 3.2.1 instead of Unreleased.

tests/utils.py
fmigneault: New one is good.

CHANGES.rst
```diff
@@ -21,10 +21,15 @@ Fixes:
 
 Changes:
 --------
-- No change.
+- Updated the format of the URL for the job results. When the prefix ``/`` was not present, 
+  URL was incorrectly handled by not prepending the required base URL location.
```
fmigneault, line 25: remove
it is duplicated in fixes

fmigneault (commented):
last fix

fmigneault (approved):

PR #272, opened 2021-07-05
 Fixed  representation according to ogcapi (fixes #263)

### LINKED ISSUE #237: [Feature] Limit query validation and integration
Jobs already support `limit` parameter to provide a subset of items. 
It must be validated against official requirement definition. 
Furthermore, other listings (eg: processes) should also support it. 

[requirements/collections/REQ_rc-limit-collections-response.adoc](https://github.com/opengeospatial/ogcapi-common/blob/Working/collections/requirements/collections/REQ_rc-limit-collections-response.adoc)
[requirements/collections/REQ_rc-limit-collections-unsupported.adoc](https://github.com/opengeospatial/ogcapi-common/blob/Working/collections/requirements/collections/REQ_rc-limit-collections-unsupported.adoc)
[requirements/collections/REQ_rc-limit-definition.adoc](https://github.com/opengeospatial/ogcapi-common/blob/Working/collections/requirements/collections/REQ_rc-limit-definition.adoc)
[requirements/collections/REQ_rc-limit-items-unsupported.adoc](https://github.com/opengeospatial/ogcapi-common/blob/Working/collections/requirements/collections/REQ_rc-limit-items-unsupported.adoc)
[requirements/collections/REQ_rc-limit-response.adoc](https://github.com/opengeospatial/ogcapi-common/blob/Working/collections/requirements/collections/REQ_rc-limit-response.adoc)

### LINKED ISSUE #263: [Feature] minOccurs/maxOccurs representation
As per https://github.com/opengeospatial/ogcapi-processes/blob/master/core/openapi/schemas/inputDescription.yaml

`minOccurs` should always be an `int`
`maxOccurs` should always be an `int` or explicitly `"unbounded"`

Currently, the values are always returned as string. 

We must return the process description according to the reference schema. 
Because of pre-existing/deployed/remote processes, the conversion should be done in place such that values stored in db or read from remote instance are converted to `int` as needed before returning the process description. 

During deployment, current behaviour that handles both numerical `str` or `int` should remain valid such that it doesn't break existing request contents. Future deployments should transparently convert the items as applicable.

trapsidanadir:
Fixing representation of `maxoccurs` and `minoccurs` according to ogcapi (fixes #263).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,6 +27,8 @@ Fixes:
 ------
 - Fix backward compatibility of pre-deployed processes that did not define ``jobControlOptions`` that is now required.
   Missing definition are substituted in-place by default ``["execute-async"]`` mode.
+- Fix ``minOccurs`` and ``maxOccurs`` representation according to `OGC-API`
+  (fixes `#263  <https://github.com/crim-ca/weaver/issues/263>`_).
 
 `3.2.0 <https://github.com/crim-ca/weaver/tree/3.2.0>`_ (2021-06-08)
 ========================================================================
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,7 +14,8 @@ Changes:
 
 Fixes:
 ------
-- No change.
+- Fix ``minOccurs`` and ``maxOccurs`` representation according to `OGC-API`
+  (fixes `#263  <https://github.com/crim-ca/weaver/issues/263>`_).
 
 `3.2.1 <https://github.com/crim-ca/weaver/tree/3.2.1>`_ (2021-06-08)
 ========================================================================
@@ -27,8 +28,6 @@ Fixes:
 ------
 - Fix backward compatibility of pre-deployed processes that did not define ``jobControlOptions`` that is now required.
   Missing definition are substituted in-place by default ``["execute-async"]`` mode.
-- Fix ``minOccurs`` and ``maxOccurs`` representation according to `OGC-API`
-  (fixes `#263  <https://github.com/crim-ca/weaver/issues/263>`_).
 
 `3.2.0 <https://github.com/crim-ca/weaver/tree/3.2.0>`_ (2021-06-08)
 ========================================================================
--- a/weaver/datatype.py
+++ b/weaver/datatype.py
@@ -821,7 +821,15 @@ class Process(Base):
     @property
     def inputs(self):
         # type: () -> Optional[List[Dict[str, Any]]]
-        return self.get("inputs")
+        inputs = self.get("inputs")
+        for input_ in inputs:
+            maxOccurs = input_.get("maxOccurs", False)
+            minOccurs = input_.get("minOccurs", False)
+            if minOccurs:
+                input_["minOccurs"] = int(minOccurs)
+            if maxOccurs and maxOccurs != "unbounded":
+                input_["maxOccurs"] = int(maxOccurs)
+        return inputs
 
     @property
     def outputs(self):
--- a/weaver/datatype.py
+++ b/weaver/datatype.py
@@ -821,15 +821,22 @@ class Process(Base):
     @property
     def inputs(self):
         # type: () -> Optional[List[Dict[str, Any]]]
+        """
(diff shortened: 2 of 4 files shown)


weaver/datatype.py
```diff
@@ -821,7 +821,16 @@ def version(self):
...
-        return self.get("inputs")
+        inputs = self.get("inputs")
+        if inputs is not None:
+            for input_ in inputs:
+                maxOccurs = input_.get("maxOccurs", False)
+                minOccurs = input_.get("minOccurs", False)
+                if minOccurs:
+                    input_["minOccurs"] = int(minOccurs)
+                if maxOccurs and maxOccurs != "unbounded":
+                    input_["maxOccurs"] = int(maxOccurs)
```
fmigneault, line 832: `minOccurs` and `maxOccurs` are mandatory in description

Were checks added because they were missing in come cases? If so, they should use the appropriate defaults, but I would tend to believe those defaults are added during deployment when omitted.

fmigneault (changes requested):
Don't forget to update the changelog with description and link to fixed issue.

weaver/datatype.py
trapsidanadir: Good point, check are not necessary there.

CHANGES.rst
```diff
@@ -27,6 +27,8 @@ Fixes:
 ------
 - Fix backward compatibility of pre-deployed processes that did not define ``jobControlOptions`` that is now required.
   Missing definition are substituted in-place by default ``["execute-async"]`` mode.
+- Fix ``minOccurs`` and ``maxOccurs`` representation according to `OGC-API`
+  (fixes `#263  <https://github.com/crim-ca/weaver/issues/263>`_).
```
fmigneault, line 31: move to unreleased

weaver/datatype.py
```diff
@@ -821,7 +821,23 @@ def version(self):
...
     def inputs(self):
         # type: () -> Optional[List[Dict[str, Any]]]
-        return self.get("inputs")
+        """
+        According to `OGC-API`, ``maxOccurs`` and ``minOccurs`` representations should be:
+            ``maxOccurs``: ``int`` or ``unbounded``
+            ``minOccurs``: ``int``
+
+        .. note::
+        Because of pre-existing/deployed/remote processes, inputs are formated to respect the valid representation.
```
fmigneault, line 830: missing indent of the description under note

weaver/datatype.py
```diff
@@ -821,7 +821,23 @@ def version(self):
     @property
     def inputs(self):
         # type: () -> Optional[List[Dict[str, Any]]]
-        return self.get("inputs")
+        """
+        According to `OGC-API`, ``maxOccurs`` and ``minOccurs`` representations should be:
+            ``maxOccurs``: ``int`` or ``unbounded``
```
fmigneault, line 826: missing quotes of unbounded

fmigneault (changes requested):

weaver/datatype.py
```diff
@@ -821,7 +821,23 @@ def version(self):
     @property
     def inputs(self):
         # type: () -> Optional[List[Dict[str, Any]]]
-        return self.get("inputs")
+        """
+        According to `OGC-API`, ``maxOccurs`` and ``minOccurs`` representations should be:
+            ``maxOccurs``: ``int`` or ``"unbounded"``
+            ``minOccurs``: ``int``
```
fmigneault, line 827: I think you will need some `- ` to make a list. 
Renders incorrectly in the doc: 
https://pavics-weaver--272.org.readthedocs.build/en/272/autoapi/weaver/datatype/index.html#weaver.datatype.Process.inputs

fmigneault: Seems like the change in ee8a697 didn't work. 
You can try to run `make docs` to tests more quickly locally the rendering.

fmigneault (approved):

weaver/datatype.py
trapsidanadir: I added the `-` before you refreshed I think that's why

PR #274, opened 2021-07-06
Fixed cwl running environment to prevent dysfunctions for next cwl patch

trapsidanadir:
Update the `cwl` payloads that are running directly in `weaver`, to avoid any dysfunction when the security [issue](https://www.crim.ca/jira/browse/DAC-437) will be fixed.

### DIFF
--- a/tests/functional/test_wps_package.py
+++ b/tests/functional/test_wps_package.py
@@ -37,10 +37,7 @@ from weaver.formats import (
     IANA_NAMESPACE,
     get_cwl_file_format
 )
-from weaver.processes.constants import (
-    CWL_REQUIREMENT_APP_DOCKER,
-    CWL_REQUIREMENT_INIT_WORKDIR
-)
+from weaver.processes.constants import CWL_REQUIREMENT_APP_DOCKER, CWL_REQUIREMENT_INIT_WORKDIR
 from weaver.utils import get_any_value
 
 EDAM_PLAIN = EDAM_NAMESPACE + ":" + EDAM_MAPPING[CONTENT_TYPE_TEXT_PLAIN]
--- a/tests/functional/test_wps_package.py
+++ b/tests/functional/test_wps_package.py
@@ -1393,10 +1393,6 @@ class WpsPackageAppWithS3BucketTest(WpsPackageConfigBase):
                     ]
                 }
             },
-            "hints": {CWL_REQUIREMENT_APP_BUILTIN: {
-                # ensure remote files are downloaded prior to CWL execution
-                "process": self._testMethodName,
-            }},
             "inputs": [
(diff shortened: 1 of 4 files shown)


tests/test_opensearch.py
```diff
@@ -90,11 +95,13 @@ def get_dummy_payload():
...
-
-def get_opensearch_process():
-    return Process(load_json_test_file("opensearch_process.json"))
+    payload = load_json_test_file("opensearch_deploy.json")
+    payload["executionUnit"][0]["unit"]["requirements"] = {
+        CWL_REQUIREMENT_APP_DOCKER: {
+            "dockerPull": "alpine:latest"
+        }
+    }
+    return payload
```
fmigneault, line 104: Could be added to the file directly instead.

fmigneault (commented):
@trapsidanadir 
Very nice. 
If you are looking for something to do, you can try doing a draft to handle [DAC-437](https://www.crim.ca/jira/browse/DAC-437).

tests/functional/test_wps_package.py
```diff
@@ -1378,6 +1382,9 @@ def test_execute_application_package_process_with_bucket(self):
             "baseCommand": "echo",
             "arguments": ["$(runtime.outdir)"],
             "requirements": {
+                CWL_REQUIREMENT_APP_DOCKER: {
+                    "dockerPull": "alpine:latest"
+                },
```
fmigneault, line 1387: The tests flagged a legit problem. This conflicts with the below `CWL_REQUIREMENT_APP_BUILTIN` in hints.
It can only be one of the two. 
There should probably be something that explicitly blocks deployment of a "builtin" process (if not already the case?) when using the Web-API.

Otherwise, this is a hack around that works just during tests since some execution mocking is happening.

trapsidanadir: In this case, the `hints` are supposed to _"ensure remote files are downloaded prior to CWL execution"_ but the [CWL documentation](https://www.commonwl.org/v1.0/CommandLineTool.html#Requirements_and_hints) explains that the `hints` report just a warning if an error occurs. 
Should we move the section hints to the requirements?

fmigneault: The reason that it was placed in `hints` is exactly for this reason. If in `requirements`, `BuiltinRequirement` causes an error because it is unknown. 
The problem here is more about the Weaver API than CWL. It looks for `requirements`/`hints` and tries to match only 1 of the [constant process types](https://github.com/crim-ca/weaver/blob/7c23830eba030e86b12d538b38de6a49d53345b6/weaver/processes/constants.py#L20-L34), in order to figure out how to instantiate it with the corresponding class.

fmigneault (approved):

PR #354, opened 2021-11-02
Job output context + Job execution improvements

### LINKED ISSUE #110: Support Output Location
We need a way to tell Weaver where to output results in order to eventually support private results.

This will require some additional fields to the configuration specifying the `private` output variant.
For the moment, [weaver.ini](https://github.com/crim-ca/weaver/blob/master/config/weaver.ini.example) uses a global `weaver.wps_output_dir` for the `public` output. 

For `private`, we need some kind of `wps_output_base_dir` combined with the authenticated user id to build `{base}/{user}` per-user outputs. Explicit format to be defined...

Maybe consider the following: 
```
weaver.wps_output.public.dir 
weaver.wps_output.public.path
weaver.wps_output.public.url
weaver.wps_output.private.dir
weaver.wps_output.private.path
weaver.wps_output.private.url
weaver.wps_output.private.auth.username
weaver.wps_output.private.auth.password
weaver.wps_output.private.auth.endpoint
weaver.wps_output.private.auth.type
```
Where for each case `dir` still refer to a directory location and `path` a relative URL route to weaver unless overridden explicitly by corresponding `url`.
Settings defined as `public` will have to fall back checking for plain `weaver.wps_output_[dir|path|url]` for compatibility reasons.

Relates to user authentication methodology #108
Relates to configuration improvements #109

fmigneault:
# Changes

- Support contextual WPS output location using ``X-WPS-Output-Context`` header to store ``Job`` results.
  When a ``Job`` is executed by providing this header with a sub-directory, the resulting outputs of the ``Job``
  will be placed and reported under the corresponding location relative to WPS outputs (path and URL).
- Replace ``Job.execute_async`` getter/setter by simple property using more generic ``Job.execution_mode``
  for storage in database. Provide ``Job.execute_async`` and ``Job.execute_sync`` properties based on stored mode.
- Simplify ``execute_process`` function executed by `Celery` task into sub-step functions where applicable.
- Simplify forwarding of ``Job`` parameters between ``PyWPS`` service ``WorkerService.execute_job`` method
  and `Celery` task instantiating it by reusing the ``Job`` object.
- Provide corresponding ``Job`` log URL along already reported log file path to facilitate retrieval from server side.
- Avoid ``Job.progress`` updates following ``failed`` or ``dismissed`` statuses to keep track of the last real progress
  percentage that was reached when that status was set.

# Fixes

- Provide requirement for https://www.crim.ca/jira/browse/DAC-170
- Provide requirement for https://www.crim.ca/jira/browse/DAC-188
- closes #110 
  Context should be provided by external method rather than relying on Weaver which holds no knowledge about the user.

### DIFF
--- a/docs/source/processes.rst
+++ b/docs/source/processes.rst
@@ -799,10 +799,10 @@ avoid conflicts. Therefore, outputs will be available with the following locatio
     Value ``WPS_OUTPUT_URL`` in above example is resolved accordingly with ``weaver.wps_output_url``,
     ``weaver.wps_output_path`` and ``weaver.url``, as per :ref:`conf_settings` details.
 
-When submitting a :term:`Job` for execution, it is also possible to provide the ``X-WPS-Output-Context`` header.
+When submitting a :term:`Job` for execution, it is possible to provide the ``X-WPS-Output-Context`` header.
 This modifies the output location to be nested under the specified directory or sub-directories.
 
-For example, providing ``X-WPS-Output-Context: project/test-1`` will result in outputs located at
+For example, providing ``X-WPS-Output-Context: project/test-1`` will result in outputs located at:
 
 .. code-block::
 
@@ -815,8 +815,13 @@ For example, providing ``X-WPS-Output-Context: project/test-1`` will result in o
     The path also **CANNOT** start by ``/``. In such cases, an HTTP error will be immediately raised indicating
     the symbols that where rejected when detected within ``X-WPS-Output-Context`` header.
 
+If desired, parameter ``weaver.wps_output_context`` can also be defined in the :ref:`conf_settings` in order to employ
+a default directory location nested under ``weaver.wps_output_dir`` when ``X-WPS-Output-Context`` header is omitted
+from the request. By default, this parameter is not defined (empty) in order to store :term:`Job` results directly under
+the configured :ter:`WPS` output directory.
+
 .. note::
-    Header ``X-WPS-Output-Context`` is ignored when using `S3` buckets as output location since they are stored
+    Header ``X-WPS-Output-Context`` is ignored when using `S3` buckets for output location since they are stored
     individually per :term:`Job` UUID, and hold no relevant *context* location. See also :ref:`conf_s3_buckets`.
 
 .. versionadded:: 4.3.0
--- a/docs/source/processes.rst
+++ b/docs/source/processes.rst
@@ -809,10 +809,11 @@ For example, providing ``X-WPS-Output-Context: project/test-1`` will result in o
     {WPS_OUTPUT_URL}/project/test-1/{JOB_UUID}/{output.ext}
 
 .. note::
-    Values provided by ``X-WPS-Output-Context`` can only contain alphanumeric, hyphens, underscores and path separators
-    that will result in a valid directory and URL locations. The path is assumed relative to the resolved :term:`WPS`
-    output directory, and will therefore reject any ``.`` or ``..`` path references.
-    The path must also **NOT** start by ``/``.
+    Values provided by ``X-WPS-Output-Context`` can only contain alphanumeric, hyphens, underscores and path
+    separators that will result in a valid directory and URL locations. The path is assumed relative by design to be
+    resolved under the :term:`WPS` output directory, and will therefore reject any ``.`` or ``..`` path references.
+    The path also **CANNOT** start by ``/``. In such cases, an HTTP error will be immediately raised indicating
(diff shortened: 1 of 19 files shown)


docs/source/processes.rst
```diff
@@ -770,13 +778,56 @@ described in |pywps-multi-output|_.
...
+    ``weaver.wps_output_path`` and ``weaver.url``, as per :ref:`conf_settings` details.
+
+When submitting a :term:`Job` for execution, it is also possible to provide the ``X-WPS-Output-Context`` header.
+This modifies the output location to be nested under the specified directory or sub-directories.
+
+For example, providing ``X-WPS-Output-Context: project/test-1`` will result in outputs located at
+
+.. code-block::
+
+    {WPS_OUTPUT_URL}/project/test-1/{JOB_UUID}/{output.ext}
```
dbyrns, line 809: In term of settings I'm not sure that we can avoid duplicating output_path/url defintion. Given the default output directory being the root of where I can specified sub-dir, I'm stuck to get that kind of folder:

- output_dir
    - uuid1.xml
    - uuid1.log
    - uuid1_dir
    - neat_workspace_buried_in_public_default_dir
        - user1
            - user1_uuid1.xml
            - ...  
        - user2
    - uuid2.xml
    - uuid2.log
    - uuid2_dir

So I see 2 choices:

1. We define a default output url/path and a context base output url/path that allow greater customization.
2. We add a default output context such as we could store default outputs to `{WPS_OUTPUT_URL}/{DEFAULT_OUTPUT_CONEXT}/{JOB_UUID}`. This way, the public stuff can be put deeper under the root directory.

docs/source/processes.rst
```diff
@@ -770,13 +778,56 @@ described in |pywps-multi-output|_.
...
+
+.. note::
+    Values provided by ``X-WPS-Output-Context`` can only contain alphanumeric, hyphens, underscores and path separators
+    that will result in a valid directory and URL locations. The path is assumed relative to the resolved :term:`WPS`
+    output directory, and will therefore reject any ``.`` or ``..`` path references.
+    The path must also **NOT** start by ``/``.
```
dbyrns, line 815: Should be stripped properly I guess, else what? error?

dbyrns (changes requested):
The only deal breaker is the output dir setting, but should not be too hard to fix.

docs/source/processes.rst
fmigneault: What I had in mind was to preserve `{DEFAULT_OUTPUT_CONEXT}` part as non existing by default to preserve the old behaviour on servers that don't require any context. In the case of DACCS for users use case, whether the a user is logged or not, a pre-hook in Twitcher/Magpie would always define the context. When not logged, that "context user" could be plainly `anonymous` or `public`. Therefore, the structure would be: 
```
output_dir
    anonymous
        uuid1.xml
        uuid1.logl
        uuid1_dir
    user1_id        
        uuid2.xml
        uuid2.logl
        uuid2_dir
```

Alternatively, I could define a `DEFAULT_OUTPUT_CONEXT` setting that is by default empty that can be overridden in the INI file specifically in birdhouse-deploy config.

docs/source/processes.rst
fmigneault: Yes, there is a check for this kind of values and corresponding test in the PR. 
I can add the detail explicitly that an error is raised in this case in the docs.

docs/source/processes.rst
dbyrns: Given that Weaver (outside of DACCS context) offers an optional output context that can be set or not. I think it would make sense to provide that `DEFAULT_OUTPUT_CONEXT` option for deployments that would not garantee a context for each and every requests. Having an empty default value, make sense too to preserve the backward compatibility.

docs/source/processes.rst
dbyrns: Yeah I saw the error handling, but I was wondering why we cannot strip them or even handle ./ prefixes silently? But in the end having one way to do a thing is better given that it's well documented.

fmigneault: I decided to raise instead of silently handling `.` because it is for one easier (especially for cases like `path/./path/ok`), but also because I explicitly forbid `..` and starting `/` to go up the tree / forge another base dir. It felt more confusing to support `.` but not the others.

docs/source/processes.rst
fmigneault: See new commits for discussed default setting.

dbyrns (approved):

PR #355, opened 2021-11-04
improve conformance with OGC for job metadata and search

fmigneault:
- support Job `type` and `processID` filter queries (relates to #268)
- provide Job status type field (resolves #351) 
- increase OGC conformance (relates to #53, #231)

PR #356, opened 2021-11-06
Job min max duration

### LINKED ISSUE #268: [Feature] Support missing query parameters for job listing
Following query parameters must be implemented.
 
- [ ] `minDuration` and `maxDuration` + test method
  (https://github.com/opengeospatial/ogcapi-processes/blob/68cdc1d61ee3cf7b343de3b57593ac90fe79f60e/core/abstract_tests/job-list/ATS_duration-response.adoc) 

- [x] `type`
  (https://github.com/opengeospatial/ogcapi-processes/blob/68cdc1d61ee3cf7b343de3b57593ac90fe79f60e/core/requirements/job-list/REQ_type-def.adoc)
  The current requirement is only to support `"process"` (default, https://github.com/opengeospatial/ogcapi-processes/blob/68cdc1d61ee3cf7b343de3b57593ac90fe79f60e/core/requirements/job-list/REQ_type-response.adoc). For Weaver, we should also support matched against the process type (`wps`, `builtin`) and `provider`.
  Nb, contrary to `process` query, this `type=process|...` query doesn't limit to a specific ID. 
  It is similar to `group` that Weaver already offers, but doesn't change the output format by groups, only filtered by jobs that are associated to a `process` (aka: any that are not `provider`). 

- [x] `processID`
  (https://github.com/opengeospatial/ogcapi-processes/blob/68cdc1d61ee3cf7b343de3b57593ac90fe79f60e/core/requirements/job-list/REQ_processid-response.adoc)
  This will basically be a duplicate of currently supported `process`, only to be compatible. 
  Both variants `process`/`processID` will be supported for backward compatibility, making the official `processID` prioritized over `process` in case they were somehow both provided. 

All requirements listed in https://github.com/opengeospatial/ogcapi-processes/pull/226/files must also be added to the `conformsTo` list. 

introduced in https://github.com/opengeospatial/ogcapi-processes/pull/226
relates to #267

fmigneault:
# Changes

- resolve #268 
- enforce `mongo==5.0` to used improved search
- improve job search function by separating filters into distinct methods and extendable pipeline 

Depends on https://github.com/crim-ca/weaver/pull/355 for part of the job search refactoring

### DIFF
--- a/weaver/datatype.py
+++ b/weaver/datatype.py
@@ -798,7 +798,7 @@ class Job(Base):
                 updated = self.finished
             else:
                 updated = self.started
-            updated = updated or now()
+            updated = localize_datetime(updated or now())
             self.updated = updated  # apply to remain static until saved
         return localize_datetime(updated)
 
--- a/weaver/store/mongodb.py
+++ b/weaver/store/mongodb.py
@@ -725,10 +725,11 @@ class MongodbJobStore(StoreJobs, MongodbStore):
         # type: (str, Request) -> MongodbSearchFilter
         search_filters = {}
         if not request:
-            search_filters.setdefault("access", VISIBILITY_PUBLIC)
+            search_filters["access"] = VISIBILITY_PUBLIC
         else:
-            if request.has_permission("admin") and access in VISIBILITY_VALUES:
-                search_filters["access"] = access
+            if request.has_permission("admin"):
+                if access in VISIBILITY_VALUES:  # otherwise any
+                    search_filters["access"] = access
             else:
                 user_id = request.authenticated_userid
                 if user_id is not None:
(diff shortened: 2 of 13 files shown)


weaver/datatype.py
```diff
@@ -605,6 +605,20 @@ def process(self, process):
...
+        # type: () -> str
+        """
+        Obtain the type of the element associated to the creation of this job.
+
+        .. seealso::
+            - Defined in |ogc-statusInfo|
+            - Queried with https://docs.ogc.org/DRAFTS/18-062.html#_parameter_type
+        """
+        if self.service is None:
+            return "process"
```
dbyrns, line 619: I don't know how the type is used, but it looks like a variable could be helpful to compare to.

weaver/store/mongodb.py
```diff
@@ -618,82 +735,118 @@ def find_jobs(self,
...
-        elif status:
-            search_filters["status"] = status
+    @staticmethod
+    def _apply_ref_or_type_filter(job_type, process, service):
+        # type: (Optional[str], Optional[str], Optional[str]) -> MongodbSearchFilter
 
-        if notification_email is not None:
-            search_filters["notification_email"] = notification_email
+        search_filters = {}  # type: MongodbSearchFilter
+        if job_type == "process":
```
dbyrns, line 745: Found it!

dbyrns (approved):
LGTM, except for one thing and if the "public" usage is not really intended, it's not a big deal.

Issue #348, opened 2021-10-19
support dismiss job operation for nested applications in workflow


fmigneault:
## Description

As highlighted in https://github.com/crim-ca/weaver/pull/346#discussion_r731941007, the job `dismiss` operation should also forward the cancellation to sub-processes in case the cancelled execution is a `Workflow`. 

The `dismiss` call should only send the DELETE job request but not worry too much about the result, since it is not guaranteed that all ADES implement it. To validate, responses could be slightly more strict to ensure the sub-process was dismissed if the targeted ADES reports it conforms to `http://www.opengis.net/spec/ogcapi-processes-1/1.0/req/dismiss/job-dismiss-op`.

Abort operation should also be considered instead of `terminate`, as highlighted in https://github.com/crim-ca/weaver/pull/359#discussion_r750477968. This implementation https://docs.celeryproject.org/en/latest/reference/celery.contrib.abortable.html could be considered for `abort` call. 

Finally, some tests should be done regarding https://docs.celeryproject.org/en/latest/userguide/workers.html#worker-persistent-revokes to validate that this behaves as intended when the flag is added to the `worker` Docker image. The example compose should also mount an appropriate directory (or document it) where the state file will be placed for syncing task revokes.

## References

- relates to #281 
- relates to #346

PR #359, opened 2021-11-10
fix batch job dismiss missing view registrations in config

fmigneault:
# Changes

- Add the additional / corresponding endpoints for batch job dismiss operation nested respectively under a specific process and provider.

# Fixes

- Batch job dismiss request views were missing as introduced in version `4.2.0`
  The view seemed to have been detected (maybe debug env?), but this did not work when running as a deployed service.


weaver/wps_restapi/jobs/jobs.py
```diff
@@ -387,33 +390,88 @@ def get_job_status(request):
...
+
+    :param job: job to cancel or cleanup.
+    :param container:
+    :return:
+    """
+    raise_job_dismissed(job, container)
+    if job.status in status.JOB_STATUS_CATEGORIES[status.JOB_STATUS_CATEGORY_RUNNING]:
+        # signal to stop celery task. Up to it to terminate remote if any.
+        LOGGER.debug("Job [%s] dismiss operation: Canceling task [%s]", job.id, job.task_id)
+        app.control.revoke(job.task_id, terminate=True)
```
dbyrns, line 426: Did you take in consideration a persistent revoke : https://docs.celeryproject.org/en/latest/userguide/workers.html#worker-persistent-revokes ?

Also see the note about using terminate=True : https://docs.celeryproject.org/en/latest/userguide/workers.html#revoke-revoking-tasks 
It should not be used programmatically, but I'm not sure what would be the proper way to stop that task? Maybe using the `signal` parameter with a less destructive signal that could even be catched by the worker and propagate the event to an eventual ADES?

dbyrns (dismissed):
It's better than nothing so I can approve this PR. Just take a look to my comment over the use of `revoke`

weaver/wps_restapi/jobs/jobs.py
fmigneault: I wasn't even aware of persistent revoke. 
This is somewhat bothersome. Why isn't it stored in the same db as tasks 🙄
The fix should be simple enough though with an edit to the Dockerfile. 
It will be up to the server though to mount a directory in the Docker to refer to persistent revokes between restarts.

Regarding `terminate`. I'm not sure why the doc makes so much emphasis on it being as last resort, but directly above it mentions: 
> When a worker receives a revoke request it will skip executing the task, but it won’t terminate an already executing task unless the terminate option is set.

In this case, dismiss operation is supposed to stop a running operation. 
Might need to try with this: https://docs.celeryproject.org/en/latest/reference/celery.contrib.abortable.html
I have not found other alternatives looking at different sources.

As for forwarding the cancellation, in that situation I agree terminate might kill the process before the task can send the request, but at the moment this is not done anyway. I'm leaving it for later with https://github.com/crim-ca/weaver/issues/348

dbyrns: > As for forwarding the cancellation, in that situation I agree terminate might kill the process before the task can send the request, but at the moment this is not done anyway. I'm leaving it for later with #348

Yes I agreed that this is not for the current PR scope.

Issue #364, opened 2021-11-16
Incorrect file rename based on reference URL


fmigneault:
## Describe the bug

When header `Content-Disposition` with `filename` parameter is provided, the value should be employed as destination file name when fetching external references. 

This is useful notably in the case of Google Drive files where endpoints to not reflect the real file name nor its extension.

## Context 

Reported by @ChaamC

PR #368, opened 2021-11-18
Fix file content disposition

fmigneault:
## Changes 

- Add handling of ``Content-Disposition`` header providing preferred ``filename`` or ``filename*`` parameters when
  fetching file references instead of the last URL fragment employed by default
  (resolves `#364 <https://github.com/crim-ca/weaver/issues/364>`_).
- Add more security validation of the obtained file name from HTTP reference, whether generated from URL path fragment
  or other header specification.


dbyrns (approved):

ChaamC (approved):
LGTM, was able to test your branch with anti-spoofing/asv processes using gdrive files and it seems to be working now! Thanks!

PR #370, opened 2021-11-22
Private Dockers

fmigneault:
Changes:
--------
- Add support of ``X-Auth-Docker`` request header that can be specified during `Process` deployment as
  authentication token that `Weaver` can use to obtain access and retrieve the `Docker` image referenced
  by the `Application Package` (`CWL`) located on a private registry.
- Add more documentation details about sample `CWL` definitions to execute script, Python and Dockerized applications.

Fixes:
------
- Fix parsing of inputs for `OpenSearch` parameters lookup that was assuming inputs were always provided as
  listing definition, not considering possible mapping definition.
- Fix incorrect documentation section ``Package as External Execution Unit Reference`` where content was omitted
  and incorrectly anchored as following ``process-esgf-cwt`` section.

References
-----------
- Resolves https://www.crim.ca/jira/browse/DAC-208


tests/test_datatype.py
```diff
@@ -59,3 +61,90 @@ def test_process_job_control_options_resolution():
...
+            assert auth.link == docker_input, f"Testing: [{docker_input}]"
+        except (TypeError, ValueError) as exc:
+            pytest.fail(f"Unexpected failure when [{docker_input}] was expected to be valid: [{exc}]")
+    for docker_input in invalid_references:
+        try:
+            DockerAuthentication("Basic", token, docker_input)
+        except (TypeError, ValueError):
+            pass
+        else:
+            raise AssertionError(f"Testing [{docker_input}] have raised invalid format but it did not.")
```
ChaamC, line 119: Not sure if error message is incomplete here ("but it did not ..?").

ChaamC (approved):
LGTM, good job!

tests/test_datatype.py
dbyrns: -> should have raised, but it did not?

dbyrns (dismissed):
Nice job! (great regex too 😃)

tests/test_datatype.py
fmigneault: I reword. It is like @dbyrns said.

PR #383, opened 2021-12-07
resolve UUID db handling

fmigneault:
## Changes

- add db revision operations
- replace str UUID-like field in db to actual UUID object
- resolve issue parsing unspecified UUID format

## Cause

With the addition of `X-Auth-Docker` header, an UUID was stored as-is. 
With more recent versions of `pymongo`, the unspecified representation type directly raises when reading the object.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,11 +10,12 @@ Changes
 
 Changes:
 --------
-- No change.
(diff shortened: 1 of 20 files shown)


dbyrns (commented):
I performed a quick review anyway and everything looks good.

Issue #269, opened 2021-06-14
[Feature] Add missing query parameters for processes listing


fmigneault:
## Description

As introduced by https://github.com/opengeospatial/ogcapi-processes/pull/226

It should be possible to obtain a filtered list of processes using:
- [ ] `limit`
  https://github.com/opengeospatial/ogcapi-processes/blob/68cdc1d61ee3cf7b343de3b57593ac90fe79f60e/core/abstract_tests/core/ATS_process-list-limit-def.adoc
  https://github.com/opengeospatial/ogcapi-processes/blob/68cdc1d61ee3cf7b343de3b57593ac90fe79f60e/core/abstract_tests/core/ATS_process-list-limit-response.adoc
  
- [ ] `page` (our implementation method of paging mechanism, no specific requirement in OGC-API-Processes)

- [ ] Furthermore, because of the addition of those paging operations, the corresponding `links` should be returned in the response (similar to job list): `next`, `last`, `prev`, `first`, etc. 
See also #267 for corresponding queries of jobs.
These links should also be included as headers as per #378.

- [ ] Relevant conformance items must be added:  
  - `/req/core/pl-limit-definition` 
  - `/req/job-list/jl-limit-definition`

## References

- http://docs.ogc.org/DRAFTS/18-062.html#sc_limit_parameter

PR #385, opened 2021-12-08
Support process-list paging, sorting and relevant links

fmigneault:
## changes
- Add ``links`` to ``/processes`` and ``/providers/{id}/processes`` listings.
- Add ``limit``, ``page`` and ``sort`` query parameters for ``/processes`` listing.
- Add ``ignore`` parameter to ``/processes`` listing when combined with ``providers=true`` to allow the similar
  behaviour supported by ``ignore`` on ``/providers`` endpoint, to effectively ignore services that cause parsing
  errors or failure to retrieve details from the remote reference.
- Add schema validation of contents returned on ``/processes`` endpoint.

## Fixes
- Fix sorting of text fields using alphabetical case-insensitive ordering.

## References

- resolves https://github.com/crim-ca/weaver/issues/269


weaver/wps_restapi/swagger_definitions.py
```diff
@@ -3842,24 +3875,56 @@ class OkGetProviderProcessesSchema(ExtendedMappingSchema):
...
+
+
+class ProvidersProcessesCollection(ExtendedMappingSchema):
+    providers = ProviderProcessesList(missing=drop)
+
+
+class ProcessListingMetadata(ExtendedMappingSchema):
+    description = "Metadata relative to the listed processes."
+    page = ExtendedSchemaNode(Integer(), misisng=drop, default=None, validator=Range(min=0))
+    limit = ExtendedSchemaNode(Integer(), missing=drop, default=None, validator=Range(min=0))
```
dbyrns, line 3914: min should be at least 1

weaver/wps_restapi/swagger_definitions.py
```diff
@@ -2406,7 +2421,13 @@ class ProcessListing(OneOfKeywordSchema):
...
 class ProcessCollection(ExtendedMappingSchema):
     processes = ProcessListing()
-    links = LinkList(missing=drop)
+
+
+class ProcessPagingQuery(ExtendedMappingSchema):
+    sort = ProcessSortEnum(missing=drop)
+    # if page is omitted but limit provided, use reasonable zero by default
+    page = ExtendedSchemaNode(Integer(allow_string=True), missing=0, default=0, validator=Range(min=0))
+    limit = ExtendedSchemaNode(Integer(allow_string=True), missing=None, default=None, validator=Range(min=0))
```
dbyrns, line 2430: min should be at least 1

dbyrns (changes requested):
Looking good apart some parameters validation. (Should not trust user 😜)

tests/wps_restapi/test_processes.py
```diff
@@ -178,20 +184,177 @@ def test_get_processes(self):
...
         assert self.process_private.identifier not in processes_id
 
+    def test_get_processes_with_paging(self):
+        test_prefix = "test-proc-temp"
+        for i in range(10):
+            p_id = "{}-{}".format(test_prefix, i)
+            proc = self.process_private = Process(id=p_id, package={}, visibility=VISIBILITY_PUBLIC)
+            self.process_store.save_process(proc)
+        _, total = self.process_store.list_processes(total=True, visibility=VISIBILITY_PUBLIC)
+        assert 10 < total < 15, "cannot run process paging test with current number of processes"
```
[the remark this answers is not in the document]
fmigneault: During test creation, the `process_store` is cleaned up (separate mongo db). 
It only contains the processes created during this `untittest.TestCase`. 
The guards here are just in case more tests get added over time and need adjustment of this test. 
It is easy to fix, but hard to understand the cause of error without this check.

weaver/wps_restapi/swagger_definitions.py
fmigneault: Good point. This will also avoid to explicitly check for the non zero in the code since it would fail validation otherwise.

weaver/wps_restapi/processes/utils.py
```diff
@@ -0,0 +1,117 @@
...
+         "type": CONTENT_TYPE_APP_JSON, "title": "Current process listing."},
+    ])
+    if provider:
+        prov_url = proc_url.rsplit("/", 1)[0]
+        links.append({"href": prov_url, "rel": "up", "type": CONTENT_TYPE_APP_JSON, "title": "Provider description."})
+    else:
+        links.append({"href": base_url, "rel": "up", "type": CONTENT_TYPE_APP_JSON, "title": "API entrypoint."})
+
+    cur_page = paging.get("page", None)
+    per_page = paging.get("limit", None)
```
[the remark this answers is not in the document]
fmigneault: resolved via schema pre-validation

dbyrns (approved):
LGTM

PR #402, opened 2022-02-12
Vault file feature

fmigneault:
Changes:
--------
- Add `Vault` endpoints providing a secured self-hosted file storage to upload local files for execution input.
- Add ``upload`` CLI operation for uploading local files to `Vault`.
- Add CLI automatic detection of local files during ``execute`` call to upload to `Vault` and retrieve them from it
  on the remote `Weaver` instance.
- Add ``-S``/``--schema`` option to CLI ``describe`` operation.
- Add more documentation examples and references related to CLI and ``WeaverClient`` usage.
- Improve Media-Type/Content-Type guesses based on known local definitions and extensions in ``weaver.formats``.
- Extend ``PyWPS`` ``WPSRequest`` to support more authorization header forwarding for inputs that could need it.

Fixes:
------
- Fix rendering of CLI *required* arguments under the appropriate argument group section when those arguments can be
  specified using prefixed ``-`` and ``--`` optional arguments format.
- Fix CLI ``url`` parameter to be provided using ``-u`` or ``--url`` without specific argument position needed.
- Fix CLI parsing of ``File`` inputs for ``execute`` operation when provided with quotes to capture full paths.
- Fix rendering of OpenAPI variable names (``additionalParameters``) employed to represent for example ``{input-id}``
  as the key within the mapping representation of inputs/outputs. The previous notation employed was incorrectly
  interpreted as HTML tags, making them partially hidden in Swagger UI.
- Fix reload of ``DockerAuthentication`` reference from database failing due to mismatched parameter names.
- Fix invalid generation and interpretation of timezone-aware datetime between local objects and loaded from database.
  Jobs created or reported without any timezone UTC offset were assumed as UTC+00:00 although corresponding datetimes
  were generated based on the local machine timezone information. Once reloaded from database, the missing timezone
  awareness made datetime stored in ISO-8601 format to be interpreted as already localized datetime.
- Fix invalid setup of generic CLI options headers for other operations than ``dismiss``.


weaver/vault/utils.py
```diff
@@ -0,0 +1,247 @@
...
+
+    :param header: Authorization header to parse.
+    :param unique: Whether only one or multiple tokens must be retrieved.
+    :return: Found mapping of UUID to access token. If unique, UUID can be ``None``.
+    """
+    if not isinstance(header, str):
+        return {}
+    header = header.lower()
+    if unique and "," in header:
+        return {}
```
ChaamC, line 133: Instead of returning empty mappings, should the code raise errors for an invalid authorization headers?

weaver/utils.py
```diff
@@ -346,33 +368,90 @@ def repr_json(data, force_string=True, **kwargs):
...
     tz_aware_dt = dt
     if dt.tzinfo is None:
-        utc = pytz.timezone("UTC")
-        aware = utc.localize(dt)
-        timezone = pytz.timezone(tz_name)
-        tz_aware_dt = aware.astimezone(timezone)
-    else:
+        tz_aware_dt = dt.astimezone()  # guess local timezone
+    if tz_name is None:
+        return tz_aware_dt
```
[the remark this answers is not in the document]
fmigneault: Actually, it would depend if the input datetime already has the timezone information or not. 
If not (`None`), datetime package assumes it is local machine time. 
Otherwise, the original tz would remain in the returned datetime, but this is not really a problem because the important thing here is to apply tz info to make sure they can be localized properly as needed. 
There is not really any request parsing/input at the moment from the API to localize times to the user's  timezone, so it will not impact for the moment.

weaver/vault/utils.py
fmigneault: It is left to the calling method to raise the appropriate HTTP code.

ChaamC (dismissed):

docs/source/cli.rst
```diff
@@ -23,3 +54,159 @@ Following are the detail for the shell :term:`CLI` which provides the same featu
...
+.. _cli_example_upload:
+
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Upload Example
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This operation allows manual upload of a local file to the :term:`Vault`.
+
+.. note::
+    When running the :ref:`execute` operation, any detected local file reference will be automatically uploaded
```
dbyrns, line 195: So Weaver doesn't rely anymore on extern hosting?

docs/source/processes.rst
```diff
@@ -709,6 +721,7 @@ combinations.
 .. |http_scheme| replace:: ``http(s)://``
 .. |s3_scheme| replace:: ``s3://``
 .. |file_scheme| replace:: ``file://``
+.. |vault_ref| replace:: ``vault:<UUID>``
```
dbyrns, line 724: Why not using the standard URL structure with the vault scheme like this `vault://<UUID>`? That would look neat beside `https://`, `s3://` and `file://` and maybe would be easier to parse since any urlparser could extract the scheme?

docs/source/cli.rst
fmigneault: This is specific only when using the CLI. Local files that are not accessible for the remote weaver are uploaded to it. 
If an URL is provided, it will act as usual.

docs/source/processes.rst
fmigneault: Good catch. I actually changed it afterwards. It is `vault://<UUID>` currently.

tests/functional/test_cli.py
```diff
@@ -338,10 +376,73 @@ def test_execute_with_auto_monitor(self):
         #   status was periodically pooled and returned 'running' until the final 'succeeded' resumes to download.
         raise NotImplementedError
 
+    # NOTE:
+    #   For all below '<>_auto_resolve_vault' test cases, the local file referenced in the Execute request body
+    #   should be automatically handled by uploading to the Vault and forwarding the relevant X-Auth-Vault header.
```
dbyrns, line 381: I guess that we cannot validate that the file has really be uploaded and accessed with correct header since all of that occurs under the hood? I mean as far as I know the execute function could have copied the file to the output location 😊.

weaver/vault/utils.py
```diff
@@ -0,0 +1,247 @@
...
+        if vault_tokens.get(value_match) is not None:
+            return {}  # cannot pick duplicates, drop both
+        vault_tokens[value_match] = token_match
+    return vault_tokens
+
+
+def get_vault_auth(request):
+    # type: (Request) -> Tuple[AnyUUID, Optional[str]]
+    """
+    Obtain the requested file if access is granted.
```
dbyrns, line 180: I don't see any access validation... Looks like it's a copy/paste from the next function.

weaver/processes/wps_package.py
```diff
@@ -1496,19 +1513,45 @@ def make_location_input(self, input_type, input_definition):
...
-        # auto-map local if possible after security check
-        input_local_ref = map_wps_output_location(input_location, self.settings)
-        if input_local_ref:
-            resp = request_extra("HEAD", input_location, settings=self.settings)
-            if resp.status_code == 200:  # if failed, following fetch will produce the appropriate HTTP error
+        # auto-map local file if possible after security check
+        if input_scheme == "vault":
+            vault_id = urlparse(input_location).hostname
+            input_url = get_vault_url(vault_id, self.settings)
+            resp = request_extra("HEAD", input_url, settings=self.settings, headers=self.auth)
```
dbyrns, line 1520: Wouldn't be easier to check directly in the db?

dbyrns (commented):
Greate PR! Will approve once the small comments are discussed/resolved

tests/functional/test_cli.py
fmigneault: If it didn't work, the test would fail because the execution would be missing an input (either missing completely, http not found or http forbidden if missing/bad header).

fmigneault: If we wanted, we could had an explicit check that the endpoint was called with some wrapping/mock magic of some function within the upload, but overkill validation in my opinion.

weaver/cli.py
```diff
@@ -398,13 +435,75 @@ def _parse_inputs(inputs):
...
+                    ctype = get_content_type(ext)
+                fmt = get_format(ctype, default=CONTENT_TYPE_TEXT_PLAIN)
+                res = self.upload(href, content_type=fmt.mime_type, url=url)
+                if res.code != 200:
+                    return res
+                vault_href = res.body["file_href"]
+                vault_id = res.body["file_id"]
+                token = res.body["access_token"]
+                auth_tokens[vault_id] = token
+                LOGGER.info("Converted [%s] -> [%s]", file, vault_href)
```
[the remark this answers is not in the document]
fmigneault: In the context of the CLI it is intentional since the user should know which file they are themselves uploading anyway.  
It would be unintentional in other places (API side). I think I only added debug messages in those cases.

weaver/vault/utils.py
fmigneault: hehe, copy-paste error from below function. Will fix.

fmigneault:
@dbyrns 
If you got some time, can you try running `test_execute_inputs_representation_literal_schema_auto_resolve_vault`. 
It seems to fail sporadically (66fefae was ✔️ , other commits failed ❌ but in different runs (Python 3.6/3.7/3.8) ). 
I'm not able to replicate locally. I'm wondering if I'm missing something that you would spot if you get the error.

Issue #247, opened 2021-04-30
[Feature] Support Prefer header and improve sync/async execution


fmigneault:
# Context

Support the different method of controlling, specifying or proposing execution mode (sync/async). 

Following PRs proposes the modification of `mode` in the payload to use `Prefer` header [[RFC-7240 #2](https://datatracker.ietf.org/doc/html/rfc7240#section-2)].
https://github.com/opengeospatial/ogcapi-processes/pull/174 
https://github.com/opengeospatial/ogcapi-processes/pull/194

# Todo

- [x] For backward compatibility reasons, `mode` should still be supported. 
- [x] The `Prefer` header would become another way to specify it, taking precedence on `mode` as it would be the official method.
  - `Prefer: respond-async` should be mapped to `mode: async-execute`. 
  - `Prefer: wait=<x>` should be mapped to `mode: sync-execute` if supported, and timeout if wait time is reached.
  - `Prefer: wait` timeout should resume the request as if `Prefer: respond-async` was indicated.
- [x] Both `mode` and `Prefer` should be optional, with `mode: auto` resolving to `async-execute` in the case of Weaver. 
- [x] ~When `sync-execute` is not supported (or vice versa with `async`) by a process, HTTP 412 should be immediately returned with a clear message of the reason if `mode` was employed.~ Since `mode` is not a header, 412 cannot be employed as precondition header failed.
- [x] When using `Prefer` instead, when the execution mode is not supported, processing should transparently ignore the unsupported mode and switch to the supported one (`Prefer` requirement is that it MUST ignore errors [[RFC-7240 #2](https://datatracker.ietf.org/doc/html/rfc7240#section-2)]). 
- [x] When `Prefer` was provided, the `Preference-Applied` header [[RFC-7240 #3](https://datatracker.ietf.org/doc/html/rfc7240#section-3)] should be provided in the response to indicate the auto-resolution. 
- [x] To support above points, processes should define 'supported' execution modes in their `jobControlOptions`. 
  This will also depend on whether Weaver correctly supports `sync` or not. Also, some type of processes like a CWL Workflow *could* be enforced `async`-only since they are usually always longer to complete to be reasonably `sync` in a timely manner.
[...]

PR #407, opened 2022-02-22
Quoting refactor and improvements

fmigneault:
## Context

Updates related to extension: https://github.com/opengeospatial/ogcapi-processes/tree/master/extensions/quotation

## Changes

- Move estimatior portion of the quoting operation into separate files and bind them with `celery` task to allow the same kind of dispatched processing as normal execution.
- Update `Quote` data type to contain status similarly to `Job` considering dispatched/async processing.
- Define `LocalizedDateTimeProperty` for reuse avoiding issues about datetime localization.
- Update OpenAPI schemas regarding `Quote` (partial/complete) and other datetime related fields.
- Add parsing of `Prefer` header allowing sync processing (relates to #247), *not yet added for `Jobs` themselves on `processes/{id}/execution` endpoint.


weaver/quotation/estimation.py
```diff
@@ -0,0 +1,147 @@
...
+        while status != QuoteStatus.COMPLETED and abort > 0:
+            wait = wait_secs(retry)
+            retry += 1
+            resp = request_extra("GET", href)
+            if resp.status_code != 200:
+                abort -= 1
+                wait = 5
+            else:
+                body = resp.json()
+                if body.get("status") == QuoteStatus.COMPLETED:
```
dbyrns, line 80: I see a small risk here : If for some reason, you got a 200 with a quote status != COMPLETED you'll end in an infinite loop. You should add an else clause that decrease abort here too.

fmigneault: We could had a global timeout to break out. Otherwise, this is the intended behaviour (ie: loop while still running).

fmigneault: Oh, I see. It is missing the update of `status` variable to check against the top `while status != QuoteStatus.COMPLETED`.
So, yes, something to break out when FAILED happens.

dbyrns (dismissed):
LGTM

PR #408, opened 2022-02-23
allow process deployment using OGC schema (without nested 'process' f…

fmigneault:
## Changes

- Support `Process` deployment using `OGC` schema (i.e.: `Process` metadata can be provided directly under
  ``processDescription`` instead of being nested under ``processDescription.process``).
  This aligns the deployment schema with reference `OGC API - Processes: Deploy, Replace, Undeploy` extension
  (see https://github.com/opengeospatial/ogcapi-processes/blob/master/extensions/deploy_replace_undeploy/standard/openapi/schemas/ogcapppkg.yaml schema).
  The previous schema for deployment with nested ``process`` field remains supported for backward compatibility.
- Fix resolution of the ``default`` field specifier under a list of supported ``formats`` during deployment.
  For various combinations such as when ``default: True`` format is omitted, or when the default is not ordered first,
  resolved ``default`` specifically for ``outputs`` definitions would be incorrect.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,7 +23,9 @@ Changes:
 
 Fixes:
 ------
-- No change.
+- Fix resolution of the ``default`` field specifier under a list of supported ``formats`` during deployment.
+  For various combinations such as when ``default: True`` format is omitted, or when the default is not ordered first,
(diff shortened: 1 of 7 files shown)


dbyrns (approved):
LGTM

Issue #375, opened 2021-12-02
Allow omitting execute outputs


fmigneault:
## Description

- [x] Allow omitting `outputs` from the `Execute` request as equivalent to requesting all of them (current behaviour anyway). 
- [x] The entry `/req/core/process-execute-default-outputs` could then be added to `conformance` endpoint. 
- [x] ~When omitted, the default `transmissionMode: value` must be used (current behaviour, although incorrectly reported).~
  As per https://docs.ogc.org/is/18-062r2/18-062r2.html#sc_execute_response, server can support only relevant modes if wants to offer as long as the responses correspond to expected behaviour.

## References

- https://docs.ogc.org/is/18-062r2/18-062r2.html#sc_process_outputs

PR #415, opened 2022-03-15
first implementation prefer header for execute job

### LINKED ISSUE #247: [Feature] Support Prefer header and improve sync/async execution

fmigneault:
# Changes

- Support ``Prefer`` header with ``wait`` or ``respond-async`` directives to select ``Job`` execution mode either
  as *synchronous* or *asynchronous* task, according to supported ``jobControlOptions`` of the relevant ``Process``
  being executed (resolves #247).
- Increase minor version of all ``builtin`` processes that will now be executable in wither (a)synchronous modes.
- Add ``weaver.exec_sync_max_wait`` and ``weaver.quote_sync_max_wait`` settings allowing custom definition for the
  maximum duration that can be specified to wait for a `synchronous` response from task workers.
- Fix https://github.com/crim-ca/weaver/issues/375.


fmigneault:
@dbyrns 
You can review directly #419 to save yourself some duplicate changes.

dbyrns (approved):
Given that https://github.com/crim-ca/weaver/pull/419 is approved, you're good to go with this one too!

Issue #376, opened 2021-12-03
Support `response: raw` execution results


fmigneault:
## Description

Support the `response: raw` representation, which implies:

- [x] Support `response=raw` and `transmissionMode: value` for a single output returns that output directly. 
  see https://docs.ogc.org/is/18-062r2/18-062r2.html#req_core_process-execute-sync-raw-value-one

- [x] Supporting the `multipart/related` format for multiple outputs, or simply the literal raw data format for individual outputs when requested as `transmissionMode: value`.
  see https://docs.ogc.org/is/18-062r2/18-062r2.html#req_core_process-execute-sync-raw-value-multi

- [x] Reporting of those outputs should be done by `Link` headers only for requested outputs as `transmissionMode: reference`.
  see https://docs.ogc.org/is/18-062r2/18-062r2.html#req_core_process-execute-sync-raw-ref

- [x] Mix of  `transmissionMode: value` and  `transmissionMode: reference` ~should do both of the previous, with the format being  `multipart/related` or the literal media-type for *remaining* output(s) as `value` according to number of outputs not requested by `reference`.~ MUST use `multipart/related` with the contents provided for outputs requested by `value`, and `Content-Location` for outputs requested by `reference`.  
  See https://docs.ogc.org/is/18-062r2/18-062r2.html#req_core_process-execute-sync-raw-mixed-multi

- [x] All corresponding `/req/core/process-execute-[execMode]-[transmissionMode]-[nOutputs]-[response]` conformance items are to be added to `/conformance` endpoint when they get implemented.

- [x] Potential replacement of `response` in body into `return` in `Prefer` header 
  https://github.com/crim-ca/weaver/issues/414

- [ ] Ensure that, even if the job was originally submitted with any `response`/`Prefer` value, the `/jobs/{jobID}/results` endpoint can be queried with a different `?response=raw|document` or corresponding `Prefer` header to obtain the desired representation. https://github.com/crim-ca/weaver/issues/727

## References

- https://docs.ogc.org/is/18-062r2/18-062r2.html#sc_execute_response (and subsequent sections with details for each case)
- https://docs.ogc.org/is/18-062r2/18-062r2.html#_response_7 (async extra details)
- must respect default resolution of #375
- generation of the multipart contents: https://github.com/requests/toolbelt#multipartform-data-encoder
- relates to [GD-535](https://crim-ca.atlassian.net/browse/GD-535)

PR #419, opened 2022-03-17
Handle transmission mode reference

### LINKED ISSUE #247: [Feature] Support Prefer header and improve sync/async execution

### LINKED ISSUE #375: Allow omitting execute outputs

### LINKED ISSUE #377: Resolve transmissionMode handling
## Describe the bug

**IMPORTANT**: 
All the following items are only for `response: document` case.
For `response: raw` refer to #376

The `transmissionMode` defines how the outputs are returned, regardless of `async` or `sync` execution.
Following a valid successful execution, the obtained status response (directly the response to the request if `sync` or from following GetStatus request if `async`), should respond according to `transmissionMode`. 

- [x] If  `transmissionMode: value` (also the default to be used for #375), the response should report the data/href as Weaver currently does. 

- [x] Fixes must be applied for Weaver to report the correct supported `transmissionMode` in regard to `response` type and must also return the items accordingly to requested method. Many tests will need to be updated as they employ `transmissionMode: reference`, although they act as if `transmissionMode: value` was specified.

- [x] All corresponding `/req/core/process-execute-[execMode]-[transmissionMode]-[nOutputs]-[response]` conformance items are to be added to `/conformance` endpoint when they get implemented.

# References

- https://docs.ogc.org/is/18-062r2/18-062r2.html#sc_execute_response
- https://docs.ogc.org/is/18-062r2/18-062r2.html#_response_7

fmigneault:
## Changes

Many adjustments for OGC compliance.

- Everything from #415 
- Properly return the `results` directly when running in `sync` mode (instead of `status` info)
- Generate the `Link` header for outputs requested with `transmissionMode=reference` 
- Some jobs utils refactors to align with new supported features

## References

- fixes #247
- fixes #375
- relates to #376 
  (raw single `value` output and/or many `reference` working, multipart not yet added)
- fixes #377


docs/source/processes.rst
```diff
@@ -424,79 +428,185 @@ This section will first describe the basics of this request format, and after go
...
+.. _proc_exec_mode:
+
+Execution Mode
+~~~~~~~~~~~~~~~~~~~~~
+
+In order to select how to execute a :term:`Process`, either `synchronously` or `asynchronously`, the ``Prefer`` header
+should be specified. If omitted, `Weaver` defaults to `asynchronous` execution. To execute `asynchronously` explicitly,
+``Prefer: respond-async`` should be used. Otherwise, the `synchronous` execution can be requested
+with ``Prefer: wait=X`` where ``X`` is the duration in seconds to wait for a response. If no worker becomes available
+within that time, or if this value is greater than ``weaver.exec_sync_max_wait``, the :term:`Job` will resume
```
dbyrns, line 533: If `weaver.exec_sync_max_wait` is 30 seconds and I submit a job using `Prefer: wait=60`, does the job resumes asynchronously directly or after 30 seconds?
Is this mode switching covered in the spec or is it a thing specific to Weaver? I feel like if I want a sync job to complete in 30 seconds I'm not necessarily ready to handle an async job...

docs/source/processes.rst
```diff
@@ -424,79 +428,185 @@ This section will first describe the basics of this request format, and after go
...
+`asynchronously` and the response will be returned. Furthermore, `synchronous` and `asynchronous` execution of
+a :term:`Process` can only be requested for corresponding ``jobControlOptions`` it reports as supported in
+its :ref:`Process Description <proc_op_describe>`. It is important to provide the ``jobControlOptions`` parameter with
+applicable modes when :ref:`Deploying a Process <proc_op_deploy>` to allow it to run as desired. By default, `Weaver`
+will assume that deployed processes are only `asynchronous` to handle longer operations.
+
+.. versionchanged::
+    By default, every :ref:`proc_builtin` :term:`Process` can accept both modes.
+    All previously deployed processes will only allow `asynchronous` execution, as only this one was supported.
```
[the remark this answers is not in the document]
fmigneault: It's not that it "cannot" but rather that to respect OGC definitions, because `execute-async` only is defined in their process description, it "must not" run them this way. So yes, redeploying them will allow them to run in sync.

docs/source/processes.rst
fmigneault: Yes, it will switch automatically to async after 30sec.
The OGC spec is not very explicit about this, but they do say that the client must be ready to receive either sync/async regardless of what was requested. They mostly refer to `Prefer` header spec itself about its definition.
This parameter is defined as a *preference* that the service can decide to honour or not. Weaver responds with `Preference-Applied` header (as per the spec) to indicate what it choose to do.

weaver/wps_restapi/jobs/utils.py
```diff
@@ -0,0 +1,699 @@
...
+
+    :param job: Job to cancel or cleanup.
+    :param container: Application settings.
+    :return: Updated and dismissed job.
+    """
+    raise_job_dismissed(job, container)
+    if job.status in JOB_STATUS_CATEGORIES[StatusCategory.RUNNING]:
+        # signal to stop celery task. Up to it to terminate remote if any.
+        LOGGER.debug("Job [%s] dismiss operation: Canceling task [%s]", job.id, job.task_id)
+        celery_app.control.revoke(job.task_id, terminate=True)
```
dbyrns, line 670: Given that the revoke command is asynchronous, could the cleanup following it (removing logs, xml, dir) be done prematurily while the process is still spitting logs waiting to be killed?

fmigneault: That is a possibility. I haven't explored this much.
If we encounter this, we could think of a way to handle it. Probably need to run another task with a delay to do a cleanup with below checks.

dbyrns: In fact, the succeeded jobs also leave their stuff,  so a mechanism to clean that should already be in place anyway. For the typical birds dumping the result in the output directory, there was a cron job for that. Does it apply to Weaver also?

fmigneault: The results that were generated by pywps and cwltool should be all cleaned up following execution by weaver after it moved them to wps-outputs. There should be 3 items `{jobId}.log`, `{jobId}.xml` (WPS status), `{jobId}/` (output results). 

The dismiss cleanup applied here is more specific for OGC Dismiss operation that requests explicitly a cleanup of those outputs, given a user downloaded them. 

If the cron job runs inside `wps-outputs/` for each *birds*, then it should pick up weaver as well, but there is not a cron specifically within weaver.

Issue #386, opened 2021-12-15
Update celery/pymongo


fmigneault:
## Description

Packages are pinned to avoid breaking changes:
```
celery[mongodb]==4.4.2
pymongo>=3.12.0,<4
```

They should be updated (and corresponding fixes applied) to support more recent `celery 5.x` and `pymongo 4.x`.

## References

- https://pymongo.readthedocs.io/en/stable/changelog.html#changes-in-version-4-0
- https://github.com/celery/celery/blob/master/docs/whatsnew-5.2.rst#upgrading-from-celery-4x
- https://github.com/celery/celery/blob/master/docs/whatsnew-5.2.rst#dropped-support-for-python-36 **[important]**

PR #422, opened 2022-03-30
update more recent dev/test dependencies (part 1)

fmigneault:
Based on #421, but upto 10f555e that was successful.
Following commits involve a lot of refactor that should be addressed later.

Following updates are required to fix some deeply nested dependency conflicts: 
- upgrade acceptable rdflib version  https://github.com/common-workflow-language/cwltool/commit/5bdb3d3dd47d8d1b3a1685220b4b6ce0f94c055e
- https://github.com/common-workflow-language/cwltool/blame/f2c4e5a6e71d4401f8d84e0b2a4660add84996c9/requirements.txt#L3

But they also involve chaning `cwltool` from `3.0.xxxx` to `3.1.xxxx` which breaks a lot of things.

relates to
- #154
- #386

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -76,6 +76,7 @@ confidence=
 # no Warning level messages displayed, use "--disable=all --enable=classes
 # --disable=W".
 disable=C0111,missing-docstring,
+        C0206,consider-using-dict-items,
         C0302,too-many-lines,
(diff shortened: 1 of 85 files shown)


dbyrns (approved):
If #421 is causing too much trouble you can go ahead with the part 1!

PR #442, opened 2022-06-02
WeaverClient/CLI with auth handler and other CLI adjustments

### LINKED ISSUE #400: Support authentication using CLI and WeaverClient
## Description

Some useful authentication handlers such as https://github.com/Ouranosinc/requests-magpie can be provided to `auth` parameter of requests to automatically resolve the authentication method before submitting a request. 

Weaver should provide a way to support this kind of handlers when targeting a server that requires authentication.

For Python `WeaverClient`, this is straightforward. Simply expose an `auth` parameter that is passed down to any request operation it accomplishes. The user will then be able to call the client with its own implementation.

For the CLI, allowing custom auth handler implementations is a slightly trickier.
Parameters should be defined in a way to allow runtime instantiation of the auth handler implementation.
For example, https://github.com/Ouranosinc/requests-magpie, would be referenced as such: 
```bash
weaver <operation> --auth-impl "requests_magpie.MagpieAuth" --auth-args "<magpie-url>,<magpie-user>,<magpie-pwd>"
```
When detecting those arguments, the CLI should instantiate the auth handler accordingly, and then pass it around as in the case of `WeaverClient`.

fmigneault:
Changes:
--------
- Add `CLI` *Authentication Handler* parameters and corresponding ``auth`` argument of instantiated classes for
  ``WeaverClient`` methods that allows inline request authentication and authorization resolution to access a
  protected service. Any *Authentication Handler* implementation can be used to fulfill required server functionalities.
- Replaced `CLI` option ``-t`` by ``-T`` (`Docker` token) during ``deploy`` operation to match naming convention of
  other options (resolves #400).
- Replaced `CLI` option ``-H`` by ``nH`` (``--no-headers``) and ``wH`` (``--with-headers``) to respectively
  enable or (explicitly) disable return of headers from response of the executed operation.
- Replaced `CLI` option ``-L`` by ``nL`` (``--no-links``) and ``wL`` (``--with-links``) to respectively
  enable (explicitly) or disable return of links from response of the executed operation.
- Replaced previously defined ``-H`` option by new ``-H/--header`` argument allowing insertion of explicitly provided
  request headers for relevant requests called by the executed operation.

Fixes:
------
- Fix `CLI` operations assuming valid JSON response to instead return error response content and status code.
- Fix `CLI` rendering of various optional arguments and groups when displaying help messages.

References
------
Resolves [DAC-460](https://crim-ca.atlassian.net/browse/DAC-460)

### DIFF
--- a/docs/source/cli.rst
+++ b/docs/source/cli.rst
@@ -80,6 +80,7 @@ Operations are equivalent between the :term:`CLI` and Python client.
 For each of the following examples, the client is created as follows:
 
 .. code-block:: python
+    :caption: Python
 
     client = WeaverClient(url="{WEAVER_URL}")
 
@@ -124,6 +125,7 @@ handler implementation.
 Below are examples of possible commands:
 
 .. code-block:: shell
+    :caption: Command Line
 
     weaver capabilities -u {WEAVER_URL} -aH requests_magpie.MagpieAuth -aU ${MAGPIE_URL} -aI <username> -aP <password>
 
@@ -168,12 +170,14 @@ desired :ref:`application-package`.
 The contents of below URL definition is also available in :ref:`example_app_pkg_script`.
 
 .. code-block:: shell
+    :caption: Command Line
 
     weaver deploy -u {WEAVER_URL} \
         -p docker-python-script-report \
         --cwl https://raw.githubusercontent.com/crim-ca/weaver/master/docs/examples/docker-python-script-report.cwl
 
 .. code-block:: python
+    :caption: Python
 
     client.deploy(
         process_id="docker-python-script-report",
@@ -196,10 +200,12 @@ Accomplishes the *Undeployment* request to remove a previously :ref:`Deployed <p
 from the service. Requires a `Weaver` or |ogc-api-proc-part2|_ compliant instance.
 
 .. code-block:: shell
+    :caption: Command Line
 
     weaver undeploy -u {WEAVER_URL} -p docker-python-script-report
 
 .. code-block:: python
+    :caption: Python
 
     client.undeploy("docker-python-script-report")
 
@@ -217,10 +223,12 @@ GetCapabilities Example
 Accomplishes the :ref:`GetCapabilities <proc_op_getcap>` request to obtain a list of available :term:`Process`.
 
 .. code-block:: shell
+    :caption: Command Line
 
     weaver capabilities -u {WEAVER_URL}
 
 .. code-block:: python
+    :caption: Python
 
     client.capabilities()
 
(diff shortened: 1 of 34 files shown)


francisPLT:
Went through most of it and up until now, everything looks pretty good. Also started testing the CLI directly. So far I like the changes and the inclusion of the Magpie Auth provider. Will continue tomorrow, especially the CLI, relevant documentation and live tests.

docs/source/cli.rst
```diff
@@ -66,32 +71,148 @@ CLI and Client Examples
...
+    - :py:class:`weaver.cli.BasicAuthHandler`
+    - :py:class:`weaver.cli.BearerAuthHandler`
+    - :py:class:`weaver.cli.CookieAuthHandler`
+    - |requests-magpie-auth|_
+
+.. |requests-magpie-auth| replace:: ``requests_magpie.MagpieAuth``
+.. _requests-magpie-auth: https://github.com/Ouranosinc/requests-magpie/blob/master/requests_magpie.py
+
+When using the :ref:`cli_commands`, the specific :py:class:`weaver.cli.AuthHandler` implementation to employ
+must be provided using the ``--auth-handler`` (``-aH``) argument. This can be an importable (installed) class module
```
francisPLT, line 120: Shouldn't the correct argument be `-aC AUTH_HANDLER_CLASS`?

`-aH HEADER` is for `--auth-header`, unless the typos are in the CLI itself.

fmigneault: You're right. Forgotten after `--auth-header` was added.

fmigneault: @francisPLT  Was your review completed?

francisPLT: Almost! Tested the new features on a DACCS deployment with the MagpieAuth handler, everything went well and behaved as expected, just finishing up on the docs and CLI code

francisPLT (dismissed):

docs/source/cli.rst
francisPLT: I tried capabilities, deploy, deploy( the same process, with the -D option), undeploy, execute, status and results. The auth part worked great in all of those. 

I didn't try the link related options though, but I didn't have problems with the links provided by the operations I tried.

I did have an SSL error, but I'm pretty sure it's related to the age of my instance's certificate (IaC launched test instance, switched off ssl for weaver for testing)

fmigneault: The `execute` worked on DACCS? Was the process only executable by the user used for the auth? I'm pleasantly surprised if this is the case.

francisPLT: With a minimal permission user (Group: Users), I received this error when I tried to only give write permissions for the process route : `"weaver.owsexceptions.OWSNoApplicableCode: Failed to retrieve WPS capabilities.`

I ended up having to give a `Recursive Read` permission at top level (`weaver/`) for it to work, on top of the `Recursive Write` for the specific process route (`weaver/processes/name-of-process`).

There's probably a way to configure it more tightly, but it's still a step more than running with with the admin account!

PR #452, opened 2022-07-12
fix cli docker username/password/token args resolution

fmigneault:
- Fix `CLI` not allowing expected combination of ``--username`` and ``--password`` for Docker authentication when
  deploying a `Process` that needs it to retrieve the referenced repository and image in its `CWL` definition.


francisPLT:
From a quick read, the code looks good, I'll be giving it a try tomorrow.

PR #461, opened 2022-08-23
add CLI/WeaverClient job retrieval operations

fmigneault:
Changes:
--------
- Add `CLI` and `WeaverClient` support of ``logs``, ``exceptions`` and ``statistics`` retrieval.
- Add `CLI` and `WeaverClient` support of `Job` search filtered by ``tags``, ``process`` and ``providers`` queries.
- Add `CLI` and `WeaverClient` support of `Job` search filtered by multiple ``status`` values.

Fixes:
------
- Fix ``tags`` query parameter not applied to filter `Job` search requests.

### DIFF
--- a/weaver/cli.py
+++ b/weaver/cli.py
@@ -1327,17 +1327,17 @@ class WeaverClient(object):
 
     @copy_doc(_job_info)
     def logs(self, *args, **kwargs):
-        return self._job_info(x_path="/logs", *args, **kwargs)
+        return self._job_info("/logs", *args, **kwargs)
 
     @copy_doc(_job_info)
     def exceptions(self, *args, **kwargs):
-        return self._job_info(x_path="/exceptions", *args, **kwargs)
+        return self._job_info("/exceptions", *args, **kwargs)
 
     errors = exceptions  # alias
 
     @copy_doc(_job_info)
     def statistics(self, *args, **kwargs):
-        return self._job_info(x_path="/statistics", *args, **kwargs)
+        return self._job_info("/statistics", *args, **kwargs)
 
     stats = statistics  # alias
 
--- a/weaver/cli.py
+++ b/weaver/cli.py
@@ -1326,18 +1326,18 @@ class WeaverClient(object):
         return self._parse_result(resp, with_links=with_links, with_headers=with_headers, output_format=output_format)
 
     @copy_doc(_job_info)
-    def logs(self, *args, **kwargs):
-        return self._job_info("/logs", *args, **kwargs)
+    def logs(self, **kwargs):
+        return self._job_info("/logs", **kwargs)
 
     @copy_doc(_job_info)
-    def exceptions(self, *args, **kwargs):
-        return self._job_info("/exceptions", *args, **kwargs)
+    def exceptions(self, **kwargs):
+        return self._job_info("/exceptions", **kwargs)
 
     errors = exceptions  # alias
 
     @copy_doc(_job_info)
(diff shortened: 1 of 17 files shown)


weaver/cli.py
```diff
@@ -2361,6 +2435,47 @@ def make_parser():
...
+            "Obtain the computation statistics details of a successful job using a reference UUID or URL. "
+            "If the job is not marked with succeeded status, this will return an error. "
+            "Only guaranteed by Weaver instances. Pure 'OGC API - Processes' servers might not implement this feature."
+        ),
+        formatter_class=ParagraphFormatter,
+    )
+    set_parser_sections(op_exceptions)
+    add_url_param(op_exceptions, required=False)
+    add_job_ref_param(op_exceptions)
+    add_shared_options(op_exceptions)
```
ChaamC, line 2477: Change these to use `op_statistics` instead

weaver/store/mongodb.py
```diff
@@ -925,6 +925,7 @@ def find_jobs(self,
 
         search_filters.update(self._apply_status_filter(status))
         search_filters.update(self._apply_ref_or_type_filter(job_type, process, service))
+        search_filters.update(self._apply_tags_filter(tags, request))
```
ChaamC, line 928: This seems to cause some problems too. `_apply_tags_filter` takes 1 argument, but there are two in this call here.

weaver/cli.py
```diff
@@ -1221,7 +1221,10 @@ def jobs(self,
         if sort is not None:
             query["sort"] = sort
         if isinstance(status, str) and status:
-            query["status"] = map_status(status)
+            status = status.split(",")
+        if isinstance(status, list) and status:
+            status = [map_status(_status) for _status in status]
+            query["status"] = ",".join(status)
```
ChaamC, line 1227: Is a list of statuses supported?
I tried with `-S succeeded` and `-S started` successfully, but with `-S succeeded started`, I get an error :
```
{
  "code": "JobInvalidParameter",
  "description": "Job query parameters failed validation.",
  "error": "Invalid",
  "cause": "{'status': '\"succeeded,running\" is not one of dismissed, failed, running, '\n           'accepted, succeeded'}",
  "value": {
    "status": "succeeded,running",
    "process": null,
    "provider": null
  }
}
```
It seems not to crash here, but during the actual `GET` request. Do you get the same problem?

weaver/cli.py
```diff
@@ -1173,9 +1174,12 @@ def jobs(self,
              sort=None,             # type: Optional[Sort]
              page=None,             # type: Optional[int]
              limit=None,            # type: Optional[int]
-             status=None,           # type: Optional[StatusType]
+             status=None,           # type: Optional[Union[StatusType, List[StatusType]]]
              detail=False,          # type: bool
              groups=False,          # type: bool
+             process=None,          # type: Optional[str]
+             provider=None,         # type: Optional[str]
+             tags=None,             # type: Optional[Union[str, List[str]]]
```
francisPLT, line 1182: These args are added, but they seem absent from the CLI options (as well as status)

weaver/cli.py
fmigneault: I noticed this mornign that the API validator blocked it. Working on it.

francisPLT:
Not sure how or why yet, but the last commit seems to have broken the status filtering on the jobs command.

Running `weaver jobs -p <process> -S succeeded -u ...` (tried via WeaverClient too) returns an empty array for a process that has successful jobs on my instance (and appeared on my previous tests)

PR #472, opened 2022-10-03
Fix problem with href input type for WeaverClient

francisPLT:


fmigneault:
@francisPLT 
I would like to have a test added in https://github.com/crim-ca/weaver/blob/master/tests/functional/test_cli.py to make sure this is handled properly.

weaver/cli.py
```diff
@@ -939,6 +939,8 @@ def _update_files(self, inputs, url=None):
                         "Cannot upload to vault: [%s]", file
                     )
                     continue
+                if not href.startswith("/"):  # Check to prevent upload of external references (http, https, etc.)
+                    continue
```
fmigneault, line 943: I wonder if I provide `-I input:File=relative/path/file.txt` to the CLI, will this change cause the relative file that would normally be found to fail? Would separating the previous `if "://" not in href and not os.path.isfile(href):` produce the intended use case (must be a local file that exists)?

tests/test_cli.py
```diff
@@ -212,10 +212,66 @@ def test_auth_handler_cookie():
...
+def test_href_inputs_not_uploaded_to_vault():
+    mock_result = OperationResult(False, code=500)
+
+    def mock_upload(_href, *_, **__):
+        return mock_result
+
+    inputs = {"source": {"href": "https://fake.domain.com/fakefile.zip"}}
+    with mock.patch("weaver.cli.WeaverClient.upload", side_effect=mock_upload):
+        result = WeaverClient()._update_files(inputs=inputs)
+    assert result is not mock_result
```
fmigneault, line 234: Add a message indicating explicitly that `upload` should not be called, since non-local file detection should skip this reference.

tests/test_cli.py
```diff
@@ -212,10 +212,66 @@ def test_auth_handler_cookie():
...
+def test_file_inputs_uploaded_to_vault():
+    fake_href = "/fake/href"
+    fake_id = "fake_id"
+    fake_token = "fake_token"
+
+    output_body = {"file_href": fake_href, "file_id": fake_id, "access_token": fake_token}
+    expected_output = ({
+                           "source": {
+                               "format": {
+                                   "mediaType": "application/zip"},
```
fmigneault, line 248: Replace by `ContentType` enum attribute.

tests/test_cli.py
```diff
@@ -212,10 +212,66 @@ def test_auth_handler_cookie():
...
+
+
+@pytest.mark.cli
+def test_file_inputs_uploaded_to_vault():
+    fake_href = "/fake/href"
+    fake_id = "fake_id"
+    fake_token = "fake_token"
+
+    output_body = {"file_href": fake_href, "file_id": fake_id, "access_token": fake_token}
+    expected_output = ({
```
fmigneault, line 245: Move the `dict` definition on newline to indent only once. Closing parenthesis aligned with `expected_output`.

tests/test_cli.py
```diff
@@ -212,10 +212,66 @@ def test_auth_handler_cookie():
...
+    inputs = {"source": {"href": "https://fake.domain.com/fakefile.zip"}}
+    with mock.patch("weaver.cli.WeaverClient.upload", side_effect=mock_upload):
+        result = WeaverClient()._update_files(inputs=inputs)
+    assert result is not mock_result
+    assert result == (inputs, {})
+
+
+@pytest.mark.cli
+def test_file_inputs_uploaded_to_vault():
+    fake_href = "/fake/href"
```
fmigneault, line 240: Just for consistency, this should be a full reference `https://some-host.com/...`.

tests/test_cli.py
```diff
@@ -212,10 +212,66 @@ def test_auth_handler_cookie():
...
+    mock_result = OperationResult(False, code=500)  # Simulate a problem with vault upload
+
+    def mock_upload(_href, *_, **__):
+        return mock_result
+
+    with tempfile.NamedTemporaryFile(mode="w", suffix=".zip") as input_file:
+        inputs = {"source": {"href": input_file.name}}
+        with mock.patch("weaver.cli.WeaverClient.upload", side_effect=mock_upload):
+            result = WeaverClient()._update_files(inputs=inputs)
+    assert result is mock_result
```
fmigneault, line 277: Add a message indicating why the error is expected in that specific case.

fmigneault (changes requested):
Please undo the extra indents added in the multi-line assert and removed spaces before comments that were visually aligned (changes not relevant to this PR).

weaver/cli.py
francisPLT: Interesting. I'll try it out.

I tried separating the previous `if` while removing this one, and it caused my unit tests to fail (those that feed an input with NamedTemporaryFiles paths).

In the meantime, I tried this and so far it seems to work with:

```
if "://" not in href and not os.path.isfile(href):
                    LOGGER.warning(
                        "Ignoring potential local file reference since it does not exist. "
                        "Cannot upload to vault: [%s]", file
                    )
                    continue
                if not os.path.isfile(href):  # Check to prevent upload of external references (http, https, etc.)
                    continue
```

fmigneault: Replace the `and` by `or` ?  Since `file://` is removed just before, if either `://` is still there or the path does not resolve, then it is not a valid local file.

francisPLT: Tested the new change we talked about and everything seemed fine with the other tests. Updated the `LOGGER.warning` to reflect the change, but I was wondering if it might not be better to change it to `info`, since it is also now an expected outcome?

fmigneault: I prefer to have it as warning since, although the outcome is expected, the problematic link was most probably added by mistake by the user. This would quickly highlight the problem and what must be fixed if the process fails afterward.

CHANGES.rst
```diff
@@ -16,7 +16,8 @@ Changes:
 
 Fixes:
 ------
-- No change.
+- Fix `cli._update_files` so there are no attempts to upload remote references to the vault.
```
fmigneault, line 19: Add a second set of quotes for RST code style.

weaver/cli.py
```diff
@@ -933,12 +933,13 @@ def _update_files(self, inputs, url=None):
                     continue
                 if href.startswith("file://"):
                     href = href[7:]
-                if "://" not in href and not os.path.isfile(href):
+                if "://" in href or not os.path.isfile(href):
                     LOGGER.warning(
-                        "Ignoring potential local file reference since it does not exist. "
+                        "Ignoring potential file reference since it probably is a remote reference or does not exist. "
```
fmigneault, line 938: I think the original "potential local file" made more sense, since the point of this code is to work only on local files. 
When something erroneous, but expected as local, is detected, then the warning is emitted. 
When files are valid remote references, there should not be a warning message since this is the expected behavior (maybe a debug log could be added though).

tests/functional/test_cli.py
```diff
@@ -1960,6 +1962,50 @@ def test_job_info_wrong_status(self):
             text = "".join(lines)
             assert expect in text
 
+    def test_execute_remote_input(self):
+        """
+        Verify that specified inputs are captured for a limited number of 1 item per ``-I`` option.
+        """
```
fmigneault, line 1968: I'm not sure to understand this description when looking at the execute call that uses only 1 `-I` argument. 
Looks like a bad copy-paste from somewhere else?

fmigneault (changes requested):
Some formatting/wording changes only.

tests/functional/test_cli.py
francisPLT: I missed that one, yes, bad copy-paste.

weaver/cli.py
francisPLT: Hmm.. ok then, but in that case we should change it back to a separate case, as was my first implementation, as the case represented by an `href` input type is not caught elsewhere in this method (and looking at https://github.com/crim-ca/weaver/blob/fix-cli-href-input/weaver/cli.py#L1037-L1043 , execute will always pass the input values through `_update_files`

This specific fix was because, using an input file containing and `href` tag, it was still uploaded to the vault with `ERROR    [weaver.cli] Execute failed. Scheme not supported for local file reference.
` as a result (why I originally used the `data` tag here : https://gitlab.com/crim.ca/clients/cubewerx/esa-snap-app/-/merge_requests/2#note_1116638381)

fmigneault: Ok. Indeed. A separate `if` would be more appropriate in that case.

francisPLT: Done. Reused the check to make sure href is not a local file

fmigneault (approved):

fmigneault:
@francisPLT Thanks!

PR #524, opened 2023-02-03
Fix `input-location` definition for `PACKAGE_DIRECTORY_TYPE` input

francisPLT:
Found out that the location input returned for `PACKAGE_DIRECTORY_TYPE` is one level too high, so this PR fixes this to return the proper directory.

The `locations` variable (https://github.com/crim-ca/weaver/blob/c237714e02dec8ed73a69d1a1d322ec435ad13c0/weaver/processes/wps_package.py#L2065) didn't contain the top-level container, only files and folders inside it, so I ended up using this approach, as it covers the possibility that the formatting of the url might change (with or without trailing "/")


francisPLT:
I'm looking into the failing test today. Looks like the created process fails during execution in that test, but haven't dug deep enough to find why and how to fix it.

CHANGES.rst
```diff
@@ -25,6 +25,9 @@ Fixes:
   combined with ``cornice`` OpenAPI schemas. Pin ``colander<2`` to avoid problems with latest changes.
 - Fix ``secure_filename`` causing valid names with leading or trailing underscores to be incorrectly unresolved
   because they get stripped out by the operation.
+- Fix ``input-location`` definition for ``PACKAGE_DIRECTORY_TYPE`` input in
+  ``weaver.processes.wps_package.WpsPackage.make_location_input``, which caused the wrong directory being given to
+  the cwl application.
```
fmigneault, line 30: annotate as: 
```
`CWL`
```

weaver/processes/wps_package.py
```diff
@@ -2069,7 +2069,8 @@ def make_location_input(self, input_type, input_definition):
                         f"Directory reference resolution method for input [{input_id}] "
                         f"from location [{input_location}] did not produce any staged file."
                     )
-                input_location = out_dir
+                input_directory_name = os.path.basename(os.path.normpath(input_definition.url))
+                input_location = os.path.join(out_dir, input_directory_name)
```
fmigneault, line 2073: There are a few tests (ex: https://github.com/crim-ca/weaver/blob/master/tests/functional/test_wps_package.py#L2240 and others below) that validate dirs submitted as input to obtain a listing. I think this change will still allow them to work, but you could add the relevant `assert` in those tests (check end of each test validating dir/value format in output) to check that the produced listing contains the last directory name.

fmigneault (changes requested):
For the failing workflow test, I think that it must be related to a similar logic that should be applied for managing the extra directory name during stage in/out in https://github.com/crim-ca/weaver/blob/master/weaver/processes/wps_process_base.py#L319-L412

CHANGES.rst
```diff
@@ -12,7 +12,9 @@ Changes
 
 Changes:
 --------
-- No change.
+- Fix `http` directory download to match implemented `AWS S3` directory download in `weaver.utils.fetch_directory`,
```
fmigneault, line 15: add double quotes for `http` and `weaver.utils.fetch_directory`

fmigneault: Revert the double quotes for ``` `AWS S3` ```. See other examples in the file. Double quotes for "code stuff" and single ones for italic emphasis on concepts.

tests/test_utils.py
```diff
@@ -634,7 +634,7 @@ def test_fetch_directory_html(include_dir_heading,       # type: bool
         out_dir = stack.enter_context(tempfile.TemporaryDirectory())
         out_files = fetch_directory(f"{tmp_host}/dir/", out_dir)
         expect_files = filter(lambda _f: _f.startswith("dir/"), test_http_dir_files)
-        expect_files = [os.path.join(out_dir, file.split("/", 1)[-1]) for file in expect_files]
+        expect_files = [os.path.join(out_dir, file) for file in expect_files]
```
fmigneault, line 637: What happens now in this case if the fetched directory was `f"{tmp_host}/` instead? 
Since there is no directory/path part in the URL, does that cause an issue with inconsistent results of directory nesting?

weaver/utils.py
```diff
@@ -2872,6 +2872,14 @@ def fetch_directory(location,                           # type: str
...
                                     settings=settings, **option_kwargs)
     elif location.startswith("http://") or location.startswith("https://"):
+        # Next two lines are added to match behavior of `download_files_s3` and replicate input directory name
+        # in output location
+        input_directory_name = os.path.basename(os.path.normpath(location))
+        out_dir = os.path.join(out_dir, input_directory_name)
+        # This is to filter out request parameters from the input location and prevent them
+        # from being used as a directory
+        if "?" in out_dir:
+            out_dir = out_dir.split("?")[0]
```
fmigneault, line 2882: In which situation did you encounter this issue. I would expect the calling function to pass only the URL part with queries and anchor fragments already removed, as the same logic must apply of all protocols.

weaver/utils.py
```diff
@@ -2872,6 +2872,14 @@ def fetch_directory(location,                           # type: str
                                     include=include, exclude=exclude,
                                     settings=settings, **option_kwargs)
     elif location.startswith("http://") or location.startswith("https://"):
+        # Next two lines are added to match behavior of `download_files_s3` and replicate input directory name
+        # in output location
+        input_directory_name = os.path.basename(os.path.normpath(location))
+        out_dir = os.path.join(out_dir, input_directory_name)
```
fmigneault, line 2878: Similar question as mentioned before. 
If the location is only "host+`/`", the full hostname will be taken as the input directory name.

ex: `https://test.com/` would create a directory `{out_dir}/test.com/`

Depending on how the URL is forged, I wonder it that could create some vulnerability?

tests/functional/test_workflow.py
```diff
@@ -1243,7 +1243,7 @@ def mock_tmp_input(requests_mock):
             )
             # check that all expected files made it through the listing/directory input/output chaining between steps
             output_files = "\n".join(os.path.join(*line.rsplit("/", 2)[-2:]) for line in output_lines)
-            expect_files = "\n".join(os.path.join("input_dir", os.path.split(file)[-1]) for file in expect_http_files)
+            expect_files = "\n".join(os.path.join("output_dir", os.path.split(file)[-1]) for file in expect_http_files)
```
fmigneault, line 1246: Just to make sure I have interpreted the reason of this change correctly. 

After passing the `merge` workflow step, the produced directory URL is something like `http://localhost/wps-outputs/{merge-jobID}/output_dir/`, with all expected files on disk under somewhere like `/tmp/wps_outputs/{merge-jobID}/output_dir/`. 

When the `listing` step receives this directory URL as input, isn't it supposed to stage it under the relevant input ID, some something like `/tmp/wps_outputs/{listing-jobID}/input_dir/output_dir/`? Originally, the last `output_dir/` part was not reused, hence the expected `input_dir` as the last part. 

If my interpretation is correct, there is probably a need to perform another check here to ensure `input_dir/output_dir/` is the full directory path. I guess it is not shown in the output text file because the `ls` call is relative to `output_dir/` only.

tests/test_utils.py
francisPLT: My understanding is that it would behave like the other comment below, `{out_dir}/test.com/`

I'll test it to make sure.

weaver/utils.py
francisPLT: It happened in this test https://github.com/crim-ca/weaver/blob/master/tests/test_utils.py#L643

It downloaded in a newly created folder named `?f=json`.

Since the function `fetch_directory` uses the location argument (https://github.com/crim-ca/weaver/blob/2c888096d318ecd3c6d67f2093de269c67c80afd/weaver/utils.py#L2877-L2878), I expect the filtering that removes the query and other fragments would be in another layer of the parsing, or somehow didn't work in the context of this test?

weaver/utils.py
francisPLT: Hmmm, there might be some vulnerability, but appart from, say `{out_dir}/../`, which would only give access to the job's directory, are there other possibilities? and besides, wouldn't it require  `https://../`, to be a valid url ? (or `https://test.com/../`)

Directory traversal would be limited because of the parsing, but I wonder what other vectors would be possible.

fmigneault: A quick tests and obtained results:

<table>
   <th>Test
   <th><code>normpath</code>
   <th><code>normpath</code>+<code>basename</code>
<tr>
<td><code>https://test.com/../</code>
<td><code>https:</code>
<td><code>https:</code>
<tr>
<td><code>https://test.com/../../</code>
<td><code>.</code>
<td><code>.</code>
<tr>
<td><code>https://test.com/../../../</code>
<td><code>..</code>
<td><code>..</code>
<tr>
<td><code>https://test.com/../../../../</code>
<td><code>../..</code>
<td><code>..</code>
</table>

So depending on the amount of `../` the following join with `out_dir` could allow someone to navigate one directory higher in the hierarchy than permitted.

weaver/utils.py
fmigneault: I see. The value obtained here:
https://github.com/crim-ca/weaver/blob/2c888096d318ecd3c6d67f2093de269c67c80afd/weaver/utils.py#L2860-L2861 could be stored in a variable and used for resolving only the directory name.
It's important to keep the query arguments though for the following calls that could need it:
https://github.com/crim-ca/weaver/blob/2c888096d318ecd3c6d67f2093de269c67c80afd/weaver/utils.py#L2865-L2866
https://github.com/crim-ca/weaver/blob/2c888096d318ecd3c6d67f2093de269c67c80afd/weaver/utils.py#L2869
https://github.com/crim-ca/weaver/blob/2c888096d318ecd3c6d67f2093de269c67c80afd/weaver/utils.py#L2884

fmigneault: You could do `loc_path = get_url_without_query(location)`, and then use `loc_path` in the current if. 
In the HTTP part, you can then work with `loc_part` directly in the code that resolves `out_dir`.

tests/functional/test_workflow.py
francisPLT: Haven't added this last assert yet, will get to it next

francisPLT: However, I looked at it quickly; what seems to happen it that `input_dir` is being replaced by `output_dir`, not adding to it. It makes sense, kind of, since the "input" is "output_dir" (https://github.com/crim-ca/weaver/blob/master/tests/functional/application-packages/WorkflowDirectoryListing/package.cwl#L20), but I don't see why it's happening yet. 

I'll have to look more into it.

weaver/utils.py
```diff
@@ -1827,6 +1828,17 @@ def get_secure_filename(file_name):
     return f"{prefix}{secure_filename(file_name)}{suffix}"
 
 
+def get_secure_directory_name(location):
```
fmigneault, line 1831: Just need the typing comment.

weaver/processes/wps_package.py
```diff
@@ -2069,7 +2069,8 @@ def make_location_input(self, input_type, input_definition):
                         f"Directory reference resolution method for input [{input_id}] "
                         f"from location [{input_location}] did not produce any staged file."
                     )
-                input_location = out_dir
+                input_directory_name = os.path.basename(os.path.normpath(input_definition.url))
```
fmigneault, line 2072: Should reuse `get_secure_directory_name`.

tests/functional/test_workflow.py
fmigneault: Yes. 
Process `merge` should produce `/tmp/wps_outputs/{merge-jobID}/output_dir/`.
With your addition, the last part `output_dir/` is preserved.
The original behaviour always nests under `<input-id>` to ensure no conflict, so `input_dir` for the `listing` process. 
The resulting fetched location is `/tmp/<cwl-stage-tmp>/input_dir/output_dir/`. 
So far, everything as expected.

With a quick test of `cwltool` with https://github.com/crim-ca/weaver/blob/master/tests/functional/application-packages/DirectoryListingProcess/package.cwl, the output contains entries like:
`/var/lib/cwl/stg9afec70c-3a16-49a3-946c-1039b5c07095/output_dir/...`

I've tried the same `listing` CWL with an extra dir input, and `cwltool` itself attributes different `stg<rand>` for each input.
Therefore, it replicates the goal of the distinct directories per `<input-id>`. 
I think it would be safe to consider this test valid as is.

fmigneault (changes requested):
Almost there!

weaver/processes/wps_package.py
francisPLT: Good spot!

fmigneault (approved):

PR #530, opened 2023-03-02
Fix `convert_input_values_schema` for multiple values

francisPLT:
Initial problem was found while trying to execute a process through the CLI, using in it's inputs an array of three or more elements all using the same input ID.

The problem was isolated to the `convert_input_values_schema` function.

Under the OGC schema, when building the list for repeated IDs, the function simply added the previous list to a new list:

```
# Old
input_prev = input_dict[input_id]
input_dict[input_id] = [input_prev, input_data]
```
Trying to add to `input_dict` a second element worked fine, but when adding a third, the result became `[[input_prev], input_date]`

Added a small check to convert `input_prev` if not already a list and switched to an append function.

Tests were adjusted to include lists of more than 2 elements and everything seems alright.

Also tested with my initial process and it also succeeded.

### DIFF
--- a/tests/processes/test_convert.py
+++ b/tests/processes/test_convert.py
@@ -1033,6 +1033,17 @@ def test_convert_input_values_schema_from_old():
         {"id": "test12", "href": "/data/file1.txt", "format": {"mediaType": "text/plain"}},
         {"id": "test12", "href": "/data/file2.txt", "format": {"mediaType": "text/plain"}},
         {"id": "test12", "href": "/data/file3.txt", "format": {"mediaType": "text/plain"}},
+        {"id": "test13", "value": "short"},
+        {"id": "test13", "value": "long"},
+        {"id": "test13", "value": "more"},
+        {"id": "test14", "value": ["val1", "val2", "val3"]},
+        {"id": "test15", "href": "https://www.somewebsite.com/dir1/", "type": "application/directory"},
+        {"id": "test15", "href": "https://www.somewebsite.com/dir2/", "type": "application/directory"},
+        {"id": "test15", "href": "https://www.somewebsite.com/dir3/", "type": "application/directory"},
+        {"id": "test16", "value": 1},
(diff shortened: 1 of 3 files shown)


tests/processes/test_convert.py
```diff
@@ -1030,6 +1030,9 @@ def test_convert_input_values_schema_from_old():
         {"id": "test10", "value": "more"},
         {"id": "test11", "href": "/data/file1.txt", "format": {"mediaType": "text/plain"}},
         {"id": "test11", "href": "/data/file2.txt", "format": {"mediaType": "text/plain"}},
+        {"id": "test12", "href": "/data/file1.txt", "format": {"mediaType": "text/plain"}},
+        {"id": "test12", "href": "/data/file2.txt", "format": {"mediaType": "text/plain"}},
+        {"id": "test12", "href": "/data/file3.txt", "format": {"mediaType": "text/plain"}},
```
fmigneault, line 1035: A few more test cases should be added to make sure.

13. \>2 literal values
14. \>2 literal values with >2 repeated IDs
15. same as 14, but with complex href

fmigneault (commented):
Nicely found!

tests/processes/test_convert.py
francisPLT: Maybe it's just the terminology, but I'm not sure I follow what the difference is between 13 and 14? Aren't the IDs already repeated? (id: test12, etc.)

For 15, complex href would be something like directory type?

fmigneault: You can have `[{"id": "testX", "value": "x"}, {"id": "testX", "value": ["x", "x"]}]`. 
The whole ID/value object repeated, or multiple values within a given object.
They were already tested, but with only 2 values instead of >2.

And for 15, I don't even remember if the code can handle
`{"id": "x", "href": [ "/path1", "/path2"], "format": "fmt"}`
(ok if it doesn't)

Good idea to test directories as well.

fmigneault (approved):

PR #540, opened 2023-05-27
investigate cause of failing CWL test only in GitHub CI

fmigneault:


fmigneault:
@fbachandcrim @f-PLT  
Found and resolved some failing tests.

Issue #18, opened 2019-03-14
Support `/results/{id}` and `/outputs/{id}` routes


fmigneault:
Only "/results" is currently supported (similar to "/outputs") due to Testbed14 conformance.
To ensure back-compatibility with similar WPS REST interfaces, these routes to request specific outputs by ID should be added.

~see: https://52north.github.io/tamis-rest-api/~

## To Do

- [x] add endpoint definitions
- [x] add conformance entries
  - e.g.: `/conf/core/job-results-async-one` as defined in [/REQ_job-results-success-async-one.adoc](https://github.com/opengeospatial/ogcapi-processes/blob/d011fe92750a0843106fcb87b6292db6e2b0fa51/core/requirements/core/REQ_job-results-success-async-one.adoc)
  - note: there are others to include as well, search for similar `job-results-[...]` terms
- [x] test that individual items can be retrieved individually as defined by the specification (consider File vs literals and which content they must support/return)
- [x] reuse/extend https://github.com/crim-ca/weaver/pull/548
  - [x] Given an `Accept` header referring to a different type than the original output format, the alternate representation (if it can be mapped), should be returned
  - [x] If the alternate format can be represented, the generated file can be stored next to the original one, with the alternate extension. It should be reused (cached) if requested again afterward.
    For example, a JSON stored under `.../{jobId}/{outputId}/file.json` could be requested in YAML, which could be stored under `.../{jobId}/{outputId}/file.yaml`. Requesting `Accept: application/x-yaml` a second time should simply return that file's contents rather than regenerating it from the JSON.
  - [x] If the alternate representation cannot be mapped (incompatible/unsupported types), raise *HTTP 406 Not Acceptable* with an appropriate message
  - [x] Support the same formatting for `../{N}` endpoint (relates to #https://github.com/crim-ca/weaver/pull/912)
- [x] Ensure `http://www.opengis.net/def/exceptions/ogcapi-processes-1/1.0/no-such-output` error for the case of invalid `outputID` from the process and/or available results in the job (see changes under https://github.com/opengeospatial/ogcapi-processes/pull/531)
- [x] Add any relevant `/req`, `/rec` and `/per` conformance definitions for `/jobs/{id}/outputs/{id}/{N}` case 
      (some requirements were modified/renamed or split up into other definitions)
      see:
  - https://github.com/opengeospatial/ogcapi-processes/pull/596
  - https://github.com/opengeospatial/ogcapi-processes/issues/583
[...]

Issue #100, opened 2020-04-16
Support the various formats that can represent a GeoTIFF


fmigneault:
- literal image/tif
- image/geotiff
- OGC WPS 1.0 came up with image/tiff;subtype=geotiff
- image/geo+tiff (similar to application/geo+json)
- all variants with tif vs tiff

PR #548, opened 2023-08-08
Job output transform

OlivGruwe:
- Add support for various GeoTIFF formats, allowing flexible handling and representation of GeoTIFFs in outputs
  (fixes `#100 <https://github.com/crim-ca/weaver/issues/100>`_).
- Add support for ``GET /results/{id}`` and `` GET /outputs/{id}`` routes to enable direct access to individual
  job result items by ID. This enhancement includes: support alternate representations based on the Accept header.
  If an alternate format (e.g., YAML for a JSON source) is requested it will be automatically generated and returned.
  Link headers containing all possible output formats, allowing retrieval via query parameters
  (e.g., output?f=application/x-yaml). (fixes `#18 <https://github.com/crim-ca/weaver/issues/18>`_).


tests/wps_restapi/test_jobs.py
```diff
@@ -1579,6 +1586,26 @@ def test_jobs_inputs_outputs_validations(self):
         with self.assertRaises(colander.Invalid):
             sd.Execute().deserialize({"outputs": {"random": {"transmissionMode": "bad"}}})
 
+    def test_job_transforms(self):
+        # First try transformation engine
+        test_transformations()
```
fmigneault, line 1591: I don't think we need to test this again here since it should already be done in [tests/transform/test_tranform.py](https://github.com/crim-ca/weaver/pull/548/files#diff-5d661929ab49da9c95d3dcdb9056cdac29f1e3762900a841183ff36c50ebfd8f)

tests/wps_restapi/test_jobs.py
```diff
@@ -1298,7 +1305,7 @@ def test_get_job_invalid_uuid(self):
         """
         # to make sure UUID is applied, use the "same format" (8-4-4-4-12), but with invalid definitions
         base_path = sd.job_service.path.format(job_id="thisisnt-some-real-uuid-allerrordata")
-        for sub_path in ["", "/inputs", "/outputs", "/results", "/logs", "exceptions"]:
+        for sub_path in ["", "/inputs", "/outputs", "/results", "/logs", "exceptions", "transforms"]:
```
fmigneault, line 1308: I think the `/` was missing from `"exceptions"` by mistake and should also be provided in `"/transforms"`.

tests/wps_restapi/test_jobs.py
```diff
@@ -1579,6 +1586,26 @@ def test_jobs_inputs_outputs_validations(self):
...
 
+    def test_job_transforms(self):
+        # First try transformation engine
+        test_transformations()
+
+        # Then try transformation in action
+        path = f"/jobs/{self.job_info[0].id}/outputs"
+        resp = self.app.get(path, headers=self.json_headers)
+        for link in resp.json["links"]:
+            header = {"Accept": link["type"], "Content-Type": link["type"]}
```
fmigneault, line 1597: The `"Accept": link["type"]` look valid, but `"Content-Type": link["type"]` doesn't make sense. 
You should remove it because the GET requests do not have a body, and therefore no `Content-Type` header applies.

Makefile
```diff
@@ -298,6 +298,28 @@ install-npm-remarklint: install-npm    ## install remark-lint dependency for 'ch
...
+#install-trfm-dependencies:   ## install required dependencies for Transformer
+#	@echo "Install required dependencies for Transformer"
+#	@bash -c '$(CONDA_CMD) pip install --upgrade -r "$(APP_ROOT)/requirements-trfm.txt"'
+#	@echo "Install with pip complete."
+
+.PHONY: install-cairo-dependencies
+install-cairo-dependencies:   ## install required dependencies for Transformer
+	@[ -f "$(shell which cairo)" ] || ( \
+		echo "Binary package manager cairo not found. Attempting to install it."; \
+		apt-get install libpangocairo-1.0-0 \
```
fmigneault, line 321: Is this absolutely necessary? Installing the `cairosvg` package does not install its dependency for you? 

I expect the Docker image will break otherwise, since this won't be installed.

Makefile
```diff
@@ -298,6 +298,28 @@ install-npm-remarklint: install-npm    ## install remark-lint dependency for 'ch
...
+#install-gdal:    ## install GDAL
+#	@echo "Installing conda packages for GDAL"
+#	@bash -c '$(CONDA_CMD) conda install -c conda-forge gdal'
+#	@echo "Install GDAL complete."
+#
+#.PHONY: install-trfm-dependencies
+#install-trfm-dependencies:   ## install required dependencies for Transformer
+#	@echo "Install required dependencies for Transformer"
+#	@bash -c '$(CONDA_CMD) pip install --upgrade -r "$(APP_ROOT)/requirements-trfm.txt"'
+#	@echo "Install with pip complete."
```
fmigneault, line 315: Leftover code to remove if not needed anymore.

Makefile
```diff
@@ -228,10 +228,10 @@ conda-env-export:		## export the conda environment
 install: install-all    ## alias for 'install-all' target
 
 .PHONY: install-run
-install-run: conda-install install-sys install-pkg install-raw 	## install requirements and application to run locally
+install-run: conda-install install-sys install-pkg install-raw install-transform ## install requirements and application to run locally
 
 .PHONY: install-all
-install-all: conda-install install-sys install-pkg install-pip install-dev  ## install application with all dependencies
+install-all: conda-install install-sys install-pkg install-pip install-transform ## install application with all dependencies
```
fmigneault, line 234: `install-dev` should remain here

weaver/tranform/transform.py
fmigneault, line 1: Add docstrings and typings.

weaver/tranform/transform.py
```diff
@@ -0,0 +1,379 @@
...
+    <head></head>
+    <body><p>%CONTENT%</p></body>
+    </html>"""
+
+
+FAMILIES = [
+    ["text/plain", "text/html", "application/pdf"],
+    ["image/png", "image/gif", "image/jpeg", "image/tiff", "image/svg+xml", "application/pdf"],
+    ["text/csv", "application/xml", "application/x-yaml", "application/json"]
+]
```
fmigneault, line 42: I think those should be defined in `weaver/formats.py` and reuse the media-type references defined in there.

weaver/wps_restapi/swagger_definitions.py
```diff
@@ -354,6 +362,13 @@ class URL(ExtendedSchemaNode):
     format = "url"
 
 
+class NRST(ExtendedSchemaNode):
+    schema_type = String
```
fmigneault, line 366: Add the title `NamespacedRelationshipType` so it gets used in the generated OpenAPI docs.

weaver/wps_restapi/swagger_definitions.py
```diff
@@ -354,6 +362,13 @@ class URL(ExtendedSchemaNode):
...
+    description = "Linked to an output pattern."
+    example = "some-object-slug-name"
```
fmigneault, line 368: I don't think this example properly represents what this schema attempts to handle.

weaver/wps_restapi/swagger_definitions.py
```diff
@@ -730,7 +745,8 @@ class LinkRelationshipType(OneOfKeywordSchema):
             "Relationship of the link to the current content. "
             "This should be one item amongst registered relations https://www.iana.org/assignments/link-relations/."
         )),
-        URL(description="Fully qualified extension link relation to the current content.")
+        URL(description="Fully qualified extension link relation to the current content."),
+        NRST(description="Link to Job output_id.") # NamespacedRelationshipType
```
fmigneault, line 749: Replace by something along the lines of "Namespaced relationship type.". 
Even if used in this case for job outputs, it is not exclusive to it in this schema.

weaver/wps_restapi/swagger_definitions.py
```diff
@@ -6727,6 +6784,25 @@ class GoneVaultFileDownloadResponse(ExtendedMappingSchema):
 get_prov_outputs_responses.update({
     "403": ForbiddenProviderLocalResponseSchema(),
 })
+
+get_job_output_responses = {
+    "200": OkGetJobOutputsResponse(description="success", examples={
+        "JobOutput": {
+            "summary": "Obtained wanted job value following process execution.",
+            "value": "Depending on media-type",
+        }
```
fmigneault, line 6793: Since no actual example is provided here, it is better to omit it instead.

weaver/wps_restapi/swagger_definitions.py
```diff
@@ -2951,6 +2982,21 @@ class ProviderJobStatisticsEndpoint(ProviderProcessPath, JobPath):
...
+
+
+class ProviderTransformerEndpoint(ProviderOutputsEndpoint):
+    deprecated = True
+    header = RequestHeaders()
+
+
+class JobTransformerEndpoint(JobPath):
+    deprecated = True
+    header = RequestHeaders()
```
fmigneault, line 2997: Should not have the `deprecated = True`. Those are new!

weaver/wps_restapi/jobs/utils.py
```diff
@@ -78,6 +80,55 @@
...
+            "alternatives": [f for f in [fam for fam in transform.FAMILIES if
+                                         o["mimeType"] in fam and "/pdf" not in o["mimeType"]]]
+        })
+    return outputs
+
+
+def get_link(output_id, mime_type, url):
+    return {
+        "href": url + "/" + output_id + "?f=" + mime_type, "rel": "output:" + output_id,
+        "type": mime_type, "title": "Link to job " + output_id + " result in " + mime_type
```
fmigneault, line 102: Reformat each field on its own line. Hard to read this way. Consider also using `f"{}"` stings.

weaver/wps_restapi/jobs/utils.py
fmigneault, line 1: Add missing typings.

weaver/processes/builtin/__init__.py
```diff
@@ -208,7 +208,7 @@ def run(self, runtime_context, **kwargs):
         try:
             self._validate_process()
             self._update_command()
-            super(BuiltinProcessJobBase, self).run(runtime_context, **kwargs)
+            super(BuiltinProcessJobBase, self).process(runtime_context, **kwargs)
```
fmigneault, line 211: Not sure about this change. Looks like an error. It should refer to the CWL's run method I believe.

weaver/wps_restapi/jobs/jobs.py
```diff
@@ -275,6 +340,24 @@ def get_job_results(request):
...
+@sd.job_transformer_service.get(tags=[sd.TAG_JOBS, sd.TAG_RESULTS], renderer=OutputFormat.JSON,
+                                schema=sd.JobTransformerEndpoint(), response_schemas=sd.get_job_transformer_responses)
+@log_unhandled_exceptions(logger=LOGGER, message=sd.InternalServerErrorResponseSchema.description)
+def get_job_transformer(request):
+    # type: (PyramidRequest) -> AnyResponseType
+    """
+    Retrieve the possible formats of an output job.
+    """
+    job = get_job(request)
+    return get_job_possible_output_formats(job)
```
fmigneault, line 358: Should do some kind of `sd.JobTranformerSchema().deserialize(<body>)` check before returning the contents to ensure they match the format of weaver/wps_restapi/examples/job_transformer.json

weaver/tranform/png2svg.py
```diff
@@ -0,0 +1,207 @@
...
+def main():
+    image = Image.open('examples/angular.png').convert('RGBA')
+    svg_image = rgba_image_to_svg_contiguous(image)
+    # svg_image = rgba_image_to_svg_pixels(image)
+    with open("examples/angular.svg", "w") as text_file:
+        text_file.write(svg_image)
+
+
+if __name__ == '__main__':
+    main()
```
fmigneault, line 207: This looks like something that should be removed.

weaver/tranform/tiff.py
fmigneault, line 1: Add missing docstrings and typings.

weaver/tranform/transform.py
```diff
@@ -0,0 +1,379 @@
...
+    csv_to_json(i, i + ".json")
+    json_to_yaml(i + ".json", o)
+
+
+class Transform:
+    def __init__(self, file_path, current_media_type: str, wanted_media_type: str):
+        self.file_path = file_path
+        self.cmt = current_media_type.lower()
+        self.wmt = wanted_media_type.lower()
+        self.output_path = self.file_path
```
fmigneault, line 272: The output path needs to be modifiable (either from init arg, or modified after before the `process`/`get` calls.
More importantly, no additional files should be added under the original WPS outputs sub-directories, because their contents could be critical to the packing/detection of other apps in a workflow chain. 
We need to consider the creation of some "sibling" directory next to the WPS-output-ID dir to avoid modifying the output checksums.

fmigneault (changes requested):
There's a few linting issues that can be resolved automatically using the `make fix[...]` targets.
Validate as well using `make check[...]` targets any other flagged issues that might need some manual adjustments. I didn't comment on all of them.

I did not go in depth into the actual implementation of the transform functions. 
I think there is a few tests missing to validate that the actual results from each transformation is valid (eg: pixel values from one format to another are properly rendered or expected text from one format to another is properly structured).

weaver/wps_restapi/jobs/jobs.py
```diff
@@ -583,11 +587,92 @@ def get_job_outputs(request):
...
+    if results:
+        result = results[0]
+    else:
+        raise HTTPNotFound(
+            json={
+                "code": "",
+                "description": "The requested output format is not in the possible output formats.",
+                "cause": "Incompatible mime Types",
+                "error": "",
+                "value": ""
```
fmigneault, line 653: Improve the details. This is misleading. The error is related to output ID not available, not the format.

weaver/wps_restapi/jobs/jobs.py
```diff
@@ -583,11 +587,92 @@ def get_job_outputs(request):
...
+            }
+        )
+    result_media_type = result["mimeType"]
```
fmigneault, line 656: use `get_field` to handle case where `mediaType` name would used instead

weaver/wps_restapi/jobs/jobs.py
```diff
@@ -583,11 +587,92 @@ def get_job_outputs(request):
...
+    result_media_type = guess_target_format(request, default=result_media_type)
+
+    # if format requested not equal to result media type and not in possible mediatypes...
+    if accept != result_media_type and accept not in possible_media_types:
```
fmigneault, line 660: no need to cross-validate accept, this is what `guess_target_format` does (+ format queries)
directly do `if result_media_type not in possible_media_types`

weaver/wps_restapi/swagger_definitions.py
fmigneault, line 1: undo the comment de-indents in this file (see diffs to find them quickly)

they are intentionally over-indented to have details aligned to facilitate reading them

fmigneault (changes requested):

tests/functional/test_wps_package.py
```diff
@@ -4402,23 +4419,25 @@ def test_execute_single_output_multipart_accept_alt_format(self):
...
+            assert line1 == line2
         outputs = self.app.get(f"/jobs/{job_id}/outputs", params={"schema": JobInputsOutputsSchema.OGC_STRICT})
         assert outputs.content_type.startswith(ContentType.APP_JSON)
         assert outputs.json["outputs"] == {
-            "output_data": "test",
             "output_json": {
-                "href": f"{out_url}/{job_id}/output_json/output.yml",
-                "type": ContentType.APP_YAML,
+                "href": f"{out_url}/{job_id}/output_json/result.json",
+                "type": ContentType.APP_JSON,
```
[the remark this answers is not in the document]
Nazim-crim: Conversion of the output/result is only applied when a specific `id` is used to retrieve the result. If there are multiple outputs, there would be no way to determine which output the `Accept `header format applies to. The` /outputs` endpoint only appends alternate link representations.

```
 {
         "id":"output_json",
         "title":"Link to job output_json result in alternate application/x-yaml",
         "hreflang":"en-CA",
         "href":"https://localhost/processes/tests-functional-test_wps_package-WpsPackageAppTestResultResponses-test_execute_single_output_multipart_accept_alt_format-EchoResultsTester/jobs/1cc09dff-0d15-4e9c-9186-b1e9f384e27b/output_json?f=application/x-yaml",
         "type":"application/x-yaml",
         "rel":"output"
      },
```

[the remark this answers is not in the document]
fmigneault: When using the multi-output endpoint, it is the `format` inside the execution request that indicates if an alternate representation is requested.

```python
{
  "outputs": {
      "output_json": {
          "transmissionMode": ExecuteTransmissionMode.VALUE,  # embed in the part contents
          "format": {"mediaType": ContentType.APP_YAML},      # request alternate output format
      }
  }
}
```

tests/test_datatype.py
```diff
@@ -215,6 +218,49 @@ def test_process_split_version(process_id, result):
...
+    # mock db functions called by offering
+    with contextlib.ExitStack() as stack:
+
+        stack.enter_context(mock.patch("weaver.processes.utils.get_settings", side_effect=_get_mocked))
+
+    process = Process(id=f"test-{uuid.uuid4()!s}", package={},
+                      outputs=[{"identifier": "output1", "formats": [{"mediaType": "image/tiff"}]}],
+                      inputs=[{"identifier": "input_1", "formats": [{"mediaType": "application/zip"}]}])
+    offer = process.offering()
+
```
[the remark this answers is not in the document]
Nazim-crim: the one on line 261?
`assert process.outputs[0]["formats"] == [{"mediaType": ContentType.IMAGE_TIFF}]`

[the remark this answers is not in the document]
fmigneault: Ah, you're right. All good.

tests/wps_restapi/test_processes.py
```diff
@@ -988,8 +991,26 @@ def deploy_process_CWL_direct(self,
...
+                    {
+                        "type": "string",
+                        "format": "binary",
+                        "contentMediaType": "application/pdf",
+                        "contentEncoding": "base64"
+                    }
+                ]
+            },
+            "formats": [{"default": True, "mediaType": "text/plain"}, {"mediaType": "text/html"},
+                        {"mediaType": "application/pdf"}]
```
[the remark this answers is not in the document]
Nazim-crim: Not sure I follow, the assert is related to the outputs for the process description following `self.deploy_process_make_visible_and_fetch_deployed(body, p_id, assert_io=False)` . In `get_cwl_docker_python_version` cwl outputs are already defined using            
```
 "outputs": {
                "output": {
                    "type": "File",
                    "outputBinding": {
                        "glob": "stdout.log"
                    },
                }
            },
```
not sure where specifying the outputs expected for the process would fit there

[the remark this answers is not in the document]
fmigneault: What I mean is that it looks like all tests after use the same `assert` against the process description contents. So, rather than having it copied many times, the `assert` could be moved inside `deploy_process_make_visible_and_fetch_deployed` to have only 1 place to maintain.

weaver/transform/transform.py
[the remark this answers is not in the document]
fmigneault: the `:type:` entries should be removed as they are redundant from the actual typing annotations

weaver/transform/tiff.py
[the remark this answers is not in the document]
fmigneault: the `:type:` entries should be removed as they are redundant from the actual typing annotations

weaver/datatype.py
```diff
@@ -2846,6 +2883,10 @@ def offering(self, schema=ProcessSchema.OGC, request=None):
             # In this situation, the lack of WPS I/O altogether requires to generate OAS from I/O merge/conversion.
             # Deployment with OAS should have generated this field already to save time or for more precise definitions.
             for io_def in process[io_type].values():
+                if io_type == "outputs":
+                    formats = io_def.get("formats", [])
+                    if formats:
+                        io_def["formats"] = extend_alternate_formats(formats, transform.CONVERSION_DICT)
```
fmigneault, line 2889: Here, I think it would be better to make an optional `conversion_dict=None` parameter in the function definition, which would use `CONVERSION_DICT` by default if no alternate dict was given. This call can then simply pass `formats` with nothing else.

This way, we can remove some of the specific logic of the transform module here to leave it in one place (the transform module), and also reduce the circular dependency issues that were encountered.

weaver/transform/tiff.py
```diff
@@ -0,0 +1,133 @@
...
+        :type index: int
+        :return: The band as a NumPy array, or None if not found.
+        :rtype: Optional[np.ndarray]
+        :raises RuntimeError: If the band index is invalid or data cannot be read.
+        """
+        if index in self.range:
+            return self.dataset.read(index)
+        return None
+
+    def get_images(self, red_band: int = 1, green_band: int = 2, blue_band: int = 3) -> List[Image.Image]:
```
fmigneault, line 106: Self-note:  "ok" for now, but will most likely make no sense when involving actual satellite geotiffs with other bands (eg: sentinel-2 RGB=(B04,B03,B02)). Maybe `rasterio` could help us directly handling/detecting common band/mission combinations to be handled automatically here...

weaver/transform/transform.py
```diff
@@ -0,0 +1,640 @@
...
+
+
+def exception_handler(func):
+    """
+    Decorator to handle exceptions in functions and log them.
+
+    :param func: Function to wrap with exception handling.
+    :type func: Callable
+    :return: The wrapped function.
+    :rtype: Callable
```
fmigneault, line 64: use typing annotations instead

weaver/transform/utils.py
fmigneault, line 1: rename the `i` parameters to `image`

weaver/wps_restapi/jobs/jobs.py
```diff
@@ -582,11 +586,97 @@ def get_job_outputs(request):
...
+    result_media_type = get_field(result, "mimeType", search_variations=True)
+    result_media_type = guess_target_format(request, default=result_media_type)
+
+    if result_media_type not in possible_media_types:
+        raise HTTPUnprocessableEntity(json={
+            "code": "InvalidMimeTypeRequested",
+            "description": "The requested output format is not in the possible output formats.",
+            "cause": "Incompatible mime Types",
+            "error": "InvalidMimeTypeRequested",
+            "value": ""
```
fmigneault, line 664: Use "Media-Types" and "MediaTypes" instead.
Pass the `result_media_type` to `value`.

weaver/wps_restapi/jobs/jobs.py
```diff
@@ -582,11 +586,97 @@ def get_job_outputs(request):
...
+    else:
+        raise HTTPNotFound(
+            json={
+                "code": "",
+                "description": "The requested output Id is not available in the job results.",
+                "cause": "The output ID is not available",
+                "error": "",
+                "value": ""
+            }
+        )
```
fmigneault, line 654: It is confusing that one branch deals with transform media-types, while the other deals with the presence of output and HTTTP error.  

Instead, do the `output_id` check against job results by itself (right after `output_id = request.matchdict.get("output_id")` is obtained), and the rest of the function can assume it is resolved.

weaver/wps_restapi/jobs/jobs.py
```diff
@@ -582,11 +586,97 @@ def get_job_outputs(request):
...
+        possible_media_types = transform.CONVERSION_DICT.get(mime_type, [])
+        possible_media_types.append(mime_type)
+    else:
+        raise HTTPNotFound(
+            json={
+                "code": "",
+                "description": "The requested output Id is not available in the job results.",
+                "cause": "The output ID is not available",
+                "error": "",
+                "value": ""
```
fmigneault, line 652: Fill the details (`output_id` in `value`, `InvalidIdentifierValue` for `code`).

Cause could indicate something more useful, such as `The output ID must be one of: {}` with the list retrieved from available IDs in `job.results`. 

Use "ID" to be consistent across messages.

weaver/wps_restapi/jobs/jobs.py
```diff
@@ -582,11 +586,97 @@ def get_job_outputs(request):
...
+            }
+        )
+    result_media_type = get_field(result, "mimeType", search_variations=True)
```
fmigneault, line 655: Should be `mime_type` to allow `search_variations` to do what it is expected.

weaver/wps_restapi/jobs/jobs.py
```diff
@@ -582,11 +586,97 @@ def get_job_outputs(request):
...
+    _, output_format = get_job_output_transmission(job, output_id, is_reference)
+    output_format = accept or output_format or result_media_type
+
+    # To ensure consistency and avoid type mismatches, all output formats are converted to a dictionary.
+    # This standardization prevents errors when handling cases in `get_job_results_single`
+    # where some formats are dictionaries and others are different types (e.g., strings or ContentType).
+    if not isinstance(output_format, dict):
+        output_format = {"mime_type": output_format}
+
+    return get_job_results_single(job, result, output_id, output_format, headers=headers, settings=settings)
```
fmigneault, line 677: This is essentially the same as the end of `get_job_results_response`: 

https://github.com/crim-ca/weaver/blob/64db3485b2dbb808a5d75e1b63a586cfa1e21147/weaver/wps_restapi/jobs/utils.py#L758-L765

Please define a function like `resolve_result_single` or similar, and invoke it in both places. 

The output transmission check should already be done by the `get_job_results_single` call, with all corresponding logic to convert data/link and transforms.

weaver/wps_restapi/jobs/jobs.py
```diff
@@ -799,27 +889,33 @@ def includeme(config):
...
+    config.add_cornice_service(sd.process_jobs_service)
+    config.add_cornice_service(sd.process_job_service)
+    config.add_cornice_service(sd.process_results_service)
+    config.add_cornice_service(sd.process_result_value_service)
+    config.add_cornice_service(sd.process_outputs_service)
+    config.add_cornice_service(sd.process_output_service)
+    config.add_cornice_service(sd.process_inputs_service)
+    config.add_cornice_service(sd.process_exceptions_service)
+    config.add_cornice_service(sd.process_logs_service)
+    config.add_cornice_service(sd.process_stats_service)
```
fmigneault, line 918: sync order with upstream branch to avoid the diff

weaver/wps_restapi/jobs/utils.py
```diff
@@ -752,12 +756,12 @@ def get_job_results_response(
...
-    # FIXME: add transform for requested output format (https://github.com/crim-ca/weaver/pull/548)
-    #   req_fmt = guess_target_format(container)   where container=request
-    #   out_fmt (see above)
-    #   out_type = result.get("type")
-    #   out_select = req_fmt or out_fmt or out_type  (resolution order/precedence)
-    out_fmt = None
+    # check accept header
+    req_fmt = (request_headers or {}).get("accept")
+    out_fmt = out_transmissions[res_id][1]
+    out_type = get_field(results[res_id], "mime_type", search_variations=True, default=None)  # a voir en debuggant
```
fmigneault, line 762: debug comment to remove

fmigneault (changes requested):
rescanned quickly the changes, many comments are still open

weaver/datatype.py
Nazim-crim: The goal was to avoid importing `CONVERSION_DICT` in the transform.utils as to not have circular depedencies. Adding a default value will cause the same issue.

tests/test_datatype.py
```diff
@@ -215,6 +218,49 @@ def test_process_split_version(process_id, result):
     assert Process.split_version(process_id) == result
 
 
+def test_process_outputs_alt():
+
+    from weaver.processes.utils import get_settings as real_get_settings
+    setup_mongodb_processstore()
```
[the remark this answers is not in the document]
Nazim-crim: without it, test fail with this failure : 
`weaver_rest_url = settings.get("weaver.wps_restapi_url")`
`AttributeError: 'NoneType' object has no attribute 'get'`

weaver/datatype.py
fmigneault: `weaver.transform.utils` already does a `from weaver.formats import ...`, and `weaver.transforms.transforms` already does `from weaver.transform.utils import ...`, so there shouldn't be an issue to have something like `weaver.transforms.const` that defines `CONVERSION_DICT` and other similar constants, and import them to define the default parameter value of `extend_alternate_formats`.

tests/test_datatype.py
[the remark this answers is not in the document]
fmigneault: It seems that problem could be introduced by the mock below, when it passes `real_get_settings(None)` explicitly. Instead, could it simply pass a relevant settings definition with the following?

```python
def _get_mocked(req=None):
    settings = {"weaver.wps_restapi_url": "http://test-weaver.com"}
    return req.registry.settings if req else real_get_settings(settings)
```

Using `setup_mongodb_processstore` could have the side effect of mongodb not accessible and failing the test, which is an error not relevant for this test.

tests/transform/test_handler.py
```diff
@@ -0,0 +1,133 @@
...
+from weaver.transform.const import CONVERSION_DICT
+from weaver.transform.handlers import Transform
+
+# Register the MIME type for .yaml files
+mimetypes.add_type("application/x-yaml", ".yaml")
+
+
+def using_mimes(func):
+    def wrapper(*args, **kwargs):
+        cmt = mimetypes.guess_type(args[0])[0]
```
fmigneault: Is this still needed since Weaver now has all YAML variants in `weaver.formats`?

Does it work if `weaver.formats.guess_target_format()` is used instead?

tests/transform/test_handler.py
```diff
@@ -0,0 +1,133 @@
...
+def test_transformations(file_name):
+    file_path = os.path.join(TRANSFORM_PATH, file_name)
+    transform(file_path)
+
+
+@pytest.mark.parametrize("file_ext,content,current_type,wanted_type", [
+    ("csv", "col1,col2\nval1,val2\n", "text/csv", "application/pdf"),
+    ("json", '{"key": "value"}', "application/json", "image/png"),
+    ("yaml", "key: value\n", "application/x-yaml", "image/png"),
+    ("xml", "<root><item>value</item></root>", "application/xml", "text/csv"),
```
fmigneault: Should use the `ContentType` definitions.

tests/transform/test_handler.py
```diff
@@ -0,0 +1,133 @@
...
+        assert isinstance(result, FileResponse)
+        assert trans.output_path == txt_file
+
+
+@pytest.mark.parametrize("file_ext,content,current_type,wanted_type,expected_ext", [
+    ("txt", "test content", "text/plain", "text/html", ".html"),
+    ("txt", "test content for PDF", "text/plain", "application/pdf", ".pdf"),
+    ("html", "<html><body><p>test content</p></body></html>", "text/html", "text/plain", ".txt"),
+    ("json", '{"key": "value", "number": 123}', "application/json", "text/plain", ".txt"),
+])
```
fmigneault, line 98: use `ContentType`

idem for a few more below

tests/wps_restapi/test_processes.py
```diff
@@ -1283,14 +1286,45 @@ def test_deploy_process_CWL_media_type_aliases_from_format(self):
...
-            }
+            },
+            {
+                "mediaType": "text/csv",
+            },
+            {
+                "mediaType": "application/xml",
+            },
+            {
+                "mediaType": "application/json",
```
fmigneault: use `ContentType`

tests/wps_restapi/test_processes.py
```diff
@@ -1386,8 +1420,29 @@ def deploy_process_CWL_direct(self,
...
+                    },
+                    {
+                        "type": "string",
+                        "contentMediaType": ContentType.TEXT_HTML
+                    },
+                    {
+                        "type": "string",
+                        "format": "binary",
+                        "contentMediaType": ContentType.APP_PDF,
+                        "contentEncoding": "base64"
```
fmigneault: use `ContentEncoding`

tests/wps_restapi/test_processes.py
```diff
@@ -1485,8 +1540,29 @@ def test_deploy_process_CWL_DockerRequirement_href(self, exec_unit_style):
...
+                        },
+                        {
+                            "type": "string",
+                            "contentMediaType": ContentType.TEXT_HTML
+                        },
+                        {
+                            "type": "string",
+                            "format": "binary",
+                            "contentMediaType": ContentType.APP_PDF,
+                            "contentEncoding": "base64"
```
fmigneault: use `ContentEncoding`

tests/wps_restapi/test_processes.py
```diff
@@ -1525,8 +1601,29 @@ def test_deploy_process_CWL_DockerRequirement_owsContext(self):
...
```
fmigneault: use `ContentEncoding`

tests/wps_restapi/test_processes.py
```diff
@@ -1563,8 +1660,29 @@ def test_deploy_process_CWL_DockerRequirement_executionUnit(self):
...
```
fmigneault: use `ContentEncoding`

tests/wps_restapi/test_processes.py
```diff
@@ -1601,8 +1719,29 @@ def test_deploy_process_CWL_DockerRequirement_executionUnit_DirectUnit(self):
...
```
fmigneault: use `ContentEncoding`

tests/wps_restapi/test_processes.py
```diff
@@ -1639,8 +1778,29 @@ def test_deploy_process_CWL_DockerRequirement_executionUnit_UnitWithMediaType(se
...
```
fmigneault: use `ContentEncoding`

tests/test_datatype.py
```diff
@@ -215,6 +217,43 @@ def test_process_split_version(process_id, result):
...
+            },
+            {
+                "mediaType": ContentType.IMAGE_SVG_XML
+            },
+            {
+                "mediaType": ContentType.APP_PDF
+            }]
+
+        # Assert that process outputs are unchanged
+        assert process.outputs[0]["formats"] == [{"mediaType": ContentType.IMAGE_TIFF}]
```
fmigneault: To make sure it is properly interpreted later, can you add explicitly "from the original database definition" to the comment.

tests/test_datatype.py
```diff
@@ -215,6 +217,43 @@ def test_process_split_version(process_id, result):
...
+        mock_settings = {"weaver.wps_restapi_url": "http://test-weaver.com"}
+        stack.enter_context(mock.patch("weaver.datatype.get_settings", return_value=mock_settings))
+
+        process = Process(id=f"test-{uuid.uuid4()!s}", package={},
+                          outputs=[{"identifier": "output1", "formats": [{"mediaType": ContentType.IMAGE_TIFF}]}],
+                          inputs=[{"identifier": "input_1", "formats": [{"mediaType": ContentType.APP_ZIP}]}])
+        offer = process.offering()
+
+        # Assert that process outputs in offering contains alternate representation
+        assert offer["outputs"]["output1"]["formats"] == [
```
fmigneault: Add a "dynamically added" to the comment to be explicit

tests/test_datatype.py
```diff
@@ -215,6 +217,43 @@ def test_process_split_version(process_id, result):
     assert Process.split_version(process_id) == result
 
 
+def test_process_outputs_alt():
+
```
fmigneault, line 223: Add a short docstring indicating:
> Validates handling of additional formats for output transform.
> 
> The output formats provided by transforms are dynamically extended from the original database process to provide alternate results. These should not modify the process definition originally deployed.

weaver/transform/handlers.py
```diff
@@ -0,0 +1,585 @@
...
+
+    return inner_function
+
+
+@exception_handler
+def image_to_any(i: str, out: str) -> None:
+    """
+    Converts image files to a specified output format. If no conversion is needed, it copies the file.
+
+    :param i: Input image file path.
```
fmigneault: adjust `i -> image_path`

same for all others below

weaver/transform/handlers.py
```diff
@@ -0,0 +1,585 @@
...
+        i = png
+
+    return images_to_any([Image.open(i)], out)
+
+
+def images_to_any(ims: List[Image.Image], out: str) -> None:
+    """
+    Processes a list of images and converts them to the desired format, saving them in the specified output path.
+
+    :param ims: List of Image objects to process.
```
fmigneault: `ims -> images`

weaver/transform/handlers.py
fmigneault, line 1: For all XML/HTML/YAML/JSON converters in this file, I think the `weaver.formats.OutputFormat` already offers the same operations.

If they indeed do, I would prefer we reuse them to avoid maintaining duplicate code features. Extend it with the missing `CSV` case so that all "text" based converters are handled with it.

They also have the shorthand names handling (eg: `json == application/json`) so it should be much simpler to invoke here.

weaver/transform/tiff.py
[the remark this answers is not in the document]
fmigneault: @Nazim-crim the `:type:` duplications are still there.

weaver/wps_restapi/jobs/utils.py
```diff
@@ -593,7 +598,7 @@ def get_results(  # pylint: disable=R1260
                 else:
                     output["dataType"] = dtype
 
-            if schema == JobInputsOutputsSchema.OGC_STRICT:
+            if strict:
```
fmigneault: This change is not valid. Removing `format` applies only for `ogc+strict`, not `old+strict` since `"old"` *requires* the `format` key.

weaver/datatype.py
```diff
@@ -1494,6 +1496,38 @@ def result_path(self, job_id=None, output_id=None, file_name=None):
...
+        Get direct links to all outputs in any possible format.
+
+        Args:
+            url (str): The base URL for constructing links.
+            results (List[Dict[str, Any]]): A list of result dictionaries containing
+                "mimeType" and "identifier".
+
+        Returns:
+            List[Dict[str, str]]: A list of dictionaries representing the links to
+                all possible output formats.
```
fmigneault: Format with the `:param <>:` and `:returns:` style.

SECURITY.md
```diff
@@ -8,9 +8,9 @@ Security and vulnerability reports should be submitted as an issue of type [Secu
...
 Patches for security vulnerabilities will be released for the following versions:
 
-| Version | Supported          |
-|---------|--------------------|
-| 4.x     | :white_check_mark: |
-| < 4.x   | :x:                |
+| Version | Supported            |
+| ------- | -------------------- |
+| 4.x     | :white\_check\_mark: |
+| < 4.x   | :x:                  |
```
fmigneault: Need to revert this. Was it caused by `make fix-md`?

CHANGES.rst
```diff
@@ -12,6 +12,13 @@ Changes
...
 Changes:
 --------
+- Add support for various GeoTIFF formats, allowing flexible handling and representation of GeoTIFFs in outputs
+  (fixes `#100 <https://github.com/crim-ca/weaver/issues/100>`_).
+- Add support for ``GET /jobs/{jobId}/results/{id}`` and ``GET /jobs/{jobId}/outputs/{id}`` routes to enable direct access to
+  individual `Job` result items by ID. This enhancement includes support alternate representations based on the ``Accept`` header.
+  If an alternate format (e.g., YAML for a JSON source) is requested it will be automatically generated and returned.
+- Return ``Link`` headers containing all possible output formats, allowing retrieval via query parameters
+  (e.g., ``/jobs/{jobId}/outputs/{id}?f=application/x-yaml``) (fixes `#18 <https://github.com/crim-ca/weaver/issues/18>`_).
 - No change.
```
fmigneault: Add indication of the `rel` used for those links.

Provide one change entry about the added `weaver.transforms` module and a brief explanation (eg: like the other comment about the test docstring) indicating the dynamically added transform formats.

Remove the "No change" entry.

weaver/utils.py
```diff
@@ -1249,6 +1249,16 @@ def get_file_header_datetime(dt):
...
 
 
+def create_content_id(first_id, second_id):
+    # type: (AnyUUID, AnyUUID) -> str
+    """
+    Generate a unique content id from passed ids.
+
+    Both ids can be strings or UUIDs.
+    """
+    return f"<{first_id}@{second_id}>"
```
fmigneault:
```suggestion
def create_content_id(resource_id, context_id):
    # type: (AnyUUID, AnyUUID) -> str
    """
    Generates a unique ``Content-ID`` from provided IDs.

    .. seealso::
        - Format of ``Content-ID`` header is defined under :rfc:`2392`.
        - Integration with ``multipart`` :term:`Media-Types` is defined under :rfc:`1521`.

    :param resource_id: Identifier of the resource unique within the context.
    :param context_id: Identifier that provides a unique reference across contexts.
    :returns: The ``Content-ID`` representation that can be directly employed as header.
    """
    return f"<{resource_id}@{context_id}>"
```

fmigneault:
@Nazim-crim 
There's a few "FIXME `https://github.com/crim-ca/weaver/issues/18`" under https://github.com/crim-ca/weaver/blob/master/weaver/wps_restapi/api.py
In most cases, it should just be a matter of uncommenting them, but we need to check if all cases of #18 are handled (eg: calling an output that wasn't requested initially must return the `http://www.opengis.net/def/exceptions/ogcapi-processes-1/1.0/no-such-output` error).

Details of the requirement: https://docs.ogc.org/DRAFTS/18-062r3.html#req_core_job-results-exception-no-such-output

SECURITY.md
Nazim-crim: yes, when running make fix i think

tests/transform/test_handler.py
Nazim-crim: yes it does

weaver/wps_restapi/jobs/utils.py
Nazim-crim: let me know if this change is correct

weaver/transform/handlers.py
Nazim-crim: Added a csv conversion, changelog updated

Nazim-crim:
@fmigneault This should cover the different comment and fix https://github.com/crim-ca/weaver/issues/18 output error code. Let me know if there's other fix to applu

fmigneault:
@Nazim-crim 
I think there is only the need to sync with https://github.com/crim-ca/weaver/pull/912 such that if certain formatting parameters are specified with the specific `N` output, the `generate_or_resolve_result()` gets called with the result of `get_job_result_by_index()`. 

For now,  `get_job_result_by_index()` assumed `return HTTPOk(json=indexed_element)` which was "OK" because `{"value": "..."}` or `{"href": "http..."}` were returned, but unless an `application/json` or `.../ogc-results` profile is requested, it should be the data itself (possibly an image for example, so `HTTP(json=...)` does not work).

Nazim-crim:
@fmigneault 

If you can check if those test case make sense with the changes  on `get_job_results_single `/ the changes in itself in the function. Not sure about the response should be wrap as default PLAIN or inside a json

fmigneault:
@Nazim-crim 
Your edits are correct. The `../{N}` must be the data itself corresponding to the `mediaType` or `type` of the output, so `text/plain` in these cases for data literals. However, we could also have a process that generates an array of JSONs. In that case, the `../{N}` would be JSON. It could be worth having a test specifically for that case because of the complex JSON nesting/parsing between the "JSON results array" and the "JSON output data".

weaver/formats.py
```diff
@@ -423,6 +428,8 @@ def convert(cls, data, to, item_root="item"):
             Unused for other representations.
         :return: Formatted output.
         """
+        from io import StringIO
+
```
fmigneault: should be at the top

or is it even needed anymore?

weaver/formats.py
```diff
@@ -448,6 +455,13 @@ def convert(cls, data, to, item_root="item"):
             if yml.endswith("\n...\n"):  # added when data is single literal or None instead of list/object
                 yml = yml[:-4]
             return yml
+        if fmt == OutputFormat.CSV:
+            import pandas as pd
+
```
fmigneault: at the top

CHANGES.rst
```diff
@@ -17,13 +17,122 @@ Changes:
...
+  and HTTP 422 when attempting to index into non-array outputs.
+
+  Added utility function ``get_job_result_by_index()`` in ``weaver.wps_restapi.jobs.utils`` to handle
+  indexed result retrieval with comprehensive validation and error handling.
 
 Fixes:
 ------
-- No change.
+- Fix ``OutputFormat.convert()`` returning ``bytes`` instead of ``str`` for XML/HTML conversions, causing write errors when
+  saving transformation results to files.
```
fmigneault: Is it still relevant? I'm not seeing the edit matching this.
Applied via the `weaver/transform/handlers.py` edits?

weaver/utils.py
```diff
@@ -1249,14 +1325,78 @@ def get_file_header_datetime(dt):
...
+def create_content_id(resource_id, context_id):
     # type: (AnyUUID, AnyUUID) -> str
     """
-    Generate a unique content id from passed ids.
+    Generates a unique ``Content-ID`` from provided IDs.
 
-    Both ids can be strings or UUIDs.
+    .. seealso::
+        - Format of ``Content-ID`` header is defined under :rfc:`2392`.
+        - Integration with ``multipart`` :term:`Media-Types` is defined under :rfc:`1521`.
```
fmigneault, line 1335: needs a newline

weaver/wps_restapi/jobs/utils.py
```diff
@@ -901,6 +945,97 @@ def get_job_results_response(
...
+    if isinstance(indexed_element, dict) and "href" in indexed_element:
+        result = indexed_element
+    else:
+        result = {"value": indexed_element}
+
+    settings = get_settings(container)
+    headers = ResponseHeaders()
+    request = container if not isinstance(container, dict) else None
+    accept = str(request.accept) if request and request.accept else None
+    return resolve_result_single(job, result, output_id, accept, headers, settings=settings)
```
fmigneault: Because this is highly dependant of the object being a request, I think we should replace the `container: AnySettingsContainer` to a `request: AnyRequestType`. Then, use `guess_target_format` directly to get the media-type. That will ensure it works with both `Accept` header and `?f=` query. The `request` object can then be passed down directly to `settings=request`. 

Can't `header={}` be passed to `resolve_result_single` to avoid the temporary var?

weaver/wps_restapi/jobs/utils.py
```diff
@@ -1130,7 +1271,16 @@ def get_job_results_single(
...
-        ctype = get_field(result, "mediaType", search_variations=True, default=ContentType.TEXT_PLAIN)
+        ctype = get_field(result, "mediaType", search_variations=True, default=None)
+
+    # If no explicit content-type and data is JSON-serializable, return as JSON
+    if not ctype and isinstance(out_data, (dict, list)):
+        return HTTPOk(json=out_data, headers=headers)
+
+    # Default to text/plain if no content-type specified
+    if not ctype:
+        ctype = ContentType.TEXT_PLAIN
```
fmigneault, line 1279: As mentioned in comment. A test of the specific JSON case should be made (if not already there?). 

For instance, an array (occurrence) of JSON-array indexed elements, or an array of JSON-objects.

weaver/wps_restapi/jobs/utils.py
```diff
@@ -591,14 +594,40 @@ def get_results(  # pylint: disable=R1260
...
+
             elif not is_ref:
                 dtype = result.get("dataType", any2wps_literal_datatype(val_data, is_value=True) or "string")
                 if ogc_api:
                     output["dataType"] = {"name": dtype}
                 else:
                     output["dataType"] = dtype
 
-            if strict:
+            if ogc_api and strict:
```
fmigneault, line 630: Fix missing in CHANGES.

tests/wps_restapi/test_jobs.py
```diff
@@ -1932,6 +1933,531 @@ def test_job_outputs_response(self):
...
+            access=Visibility.PUBLIC,
+            results=[
+                {
+                    "id": "file_array",
+                    "value": [
+                        {"href": f"{wps_out_dir}/file1.txt", "type": ContentType.TEXT_PLAIN},
+                        {"href": f"{wps_out_dir}/file2.json", "type": ContentType.APP_JSON},
+                    ]
+                },
+            ],
```
fmigneault, line 2278: Please add `type: ContentType.APP_RAW_JSON`. I think it shouldn't affect the test (well, at least currently?) but it will help indicate the behaviour how it would actually be saved in DB based on https://pavics-weaver.readthedocs.io/en/latest/processes.html#multiple-outputs

As described in that doc, this special type is used to differentiate from a "real single value" JSON. Therefore, there should another test that purposely creates a similar structure (with a JSON array of objects that happen to have `value`/`href`), which should be interpreted like a single value. In this case, doing `../{N}` should be invalid and raise the relevant HTTP error/code.

weaver/wps_restapi/jobs/utils.py
fmigneault: Nevermind. I saw an older comment about previous iterative edits. Combining like this does the same logic as what was applied before.

tests/wps_restapi/test_jobs.py
Nazim-crim: Not sure i follow, how would there be multiple href/ value for a single value ? Isn't every href pointing to a file ?
Do you have an example

like this ? 
```
            results=[
                {
                    "id": "json_output",
                    "value": [
                        {"href": "https://example.com/data1", "value": 123},
                        {"href": "https://example.com/data2", "value": 456},
                    ],
                    "mimeType": ContentType.APP_JSON,
                },
            ],
```

fmigneault: Like you propose should technically be invalid (at least, if the scheme validation is run on the result before returning it, it should raise). It can only be `oneOf[value,href]` representations.

Following is what I had in mind:

```python
results=[
    {
        "id": "json_output",
        "value": {
            "value": 123,
            "href": "https://example.com/data1"
        },
    },
]
```
or 
```python
results=[
    {
        "id": "json_output",
        "value": {
            "value": [{"href": "https://example.com/data1"}],
        },
    },
]
```
or 
```python
results=[
    {
        "id": "json_output",
        "value": [
          {
            "value": [{"href": "https://example.com/data1"}],
          }
        ],
    },
]
```

In both case, the second `"value"` nesting is not the value/href, but just a data blob that happens to have the same structure.

Nazim-crim: Got it any indexing should fail
```
 {
            "value": [{"href": "https://example.com/data1"}],
          }
```

Nazim-crim: @fmigneault Added the test for the edge case of single value json, it throw the correct exception. The test regarding the index / transform are marked as @part1 let me know if it's the right part  Thanks

Test with Array of multi json + raw json media type are also added

fmigneault: Thanks. Yes, that is the right part.

fmigneault (approved):

PR #550, opened 2023-08-31
fix import error `pywps>=4.5.2`

fmigneault:


fmigneault:
Change relevant only for `pywps>4.5.2` (ie: `main` branch https://github.com/geopython/pywps/blob/main/pywps/response/basic.py), not `=4.5.2`. 
Since the requirement is pinned to `pywps==4.5.2` for now, and that is the latest release, this is not a problem.

PR #566, opened 2023-09-22
fix nested CWL enum in array and single-value combined + CWL Workflow step output chaining with glob

fmigneault:
## Changes

- Remove attribute ``WpsProcessInterface.stage_output_id_nested`` and enforce the behavior of nesting output by ID
  under corresponding directories for all remote `Process` execution when resolving `CWL` `Workflow` steps. This
  ensures a more consistent file and directory resolution between steps of different nature (`CWL`, `WPS`, `OGC` based)
  using multiple combinations of ``glob`` patterns and expected media-types.

## Fixes

- Fix `CWL` definitions combining nested ``enum`` types as ``["null", <enum>, {type: array, items: <enum>]`` without an
  explicit ``name`` or ``SchemaDefRequirement`` causing failing ``schema_salad`` resolution under ``cwltool``. A patch
  is applied for the moment to inject a temporary ``name`` to let the `CWL` engine succeed schema validation (relates
  to https://github.com/common-workflow-language/cwltool/issues/1908.

## References

- Enum isssue relates to: 
  - https://github.com/bird-house/finch/issues/301
  - https://github.com/bird-house/finch/issues/302


fmigneault:
@perronld 
Assuming https://github.com/bird-house/finch/pull/302 is fixed (which I commented out duplicates in my local XML WPS definition to test), combined with this PR, your `ensemble_grid_point_wetdays` process should work 🤞

PR #589, opened 2023-11-15
Woraround for registry resolving when restarting celery prefork workers

perronld:
Workaround pyramid.threadlocal:get_current_registry() when restarting Celery workers processes (e.g. `--max-tasks-per-child`)


fmigneault (commented):
How does it make a workaround? 

The doc seems to indicate that `get_current_request` is just as bad as `get_current_registry`.
https://docs.pylonsproject.org/projects/pyramid/en/latest/api/threadlocal.html

Also, what happens if the code is not running with celery (ie: API or CLI), and `get_registry()` is called. Wouldn't that possibly create an endless recursion?

PR #597, opened 2024-02-06
pass session object to client

mishaschwartz:
Add option to make requests with `WeaverClient` using a `requests.Session` instance. 

This allows requests to be made with the session cookies that may already be attached to this ``requests.Session`` instance.

Useful as an alternative to authenticating with an `auth` argument when a user has already authenticated and they have access to the session cookie. This is mostly so that we can take advantage of this feature with weaver https://github.com/bird-house/birdhouse-deploy/pull/407

### DIFF
--- a/weaver/cli.py
+++ b/weaver/cli.py
@@ -328,8 +328,8 @@ class WeaverClient(object):
     monitor_interval = 5    # interval between monitor pooling job status requests
     auth = None  # type: AuthHandler
 
-    def __init__(self, url=None, auth=None):
-        # type: (Optional[str], Optional[AuthHandler]) -> None
+    def __init__(self, url=None, auth=None, session=None):
+        # type: (Optional[str], Optional[AuthHandler], Optional[Session]) -> None
         """
         Initialize the client with predefined parameters.
 
@@ -346,6 +346,7 @@ class WeaverClient(object):
             self._url = None
             LOGGER.warning("No URL provided. All operations must provide it directly or through another parameter!")
         self.auth = auth
+        self.session = session
         self._headers = {"Accept": ContentType.APP_JSON, "Content-Type": ContentType.APP_JSON}
         self._settings = {
             "weaver.request_options": {}
@@ -362,6 +363,8 @@ class WeaverClient(object):
                  ):                     # type: (...) -> AnyResponseType
         if self.auth is not None and kwargs.get("auth") is None:
             kwargs["auth"] = self.auth
+        if self.session is not None and kwargs.get("session") is None:
+            kwargs["session"] = self.session
 
         if not headers and x_headers:
             headers = x_headers
--- a/weaver/utils.py
+++ b/weaver/utils.py
@@ -1856,12 +1856,12 @@ def retry_on_cache_error(func):
     return wrapped
 
 
-def _request_call(method, url, kwargs):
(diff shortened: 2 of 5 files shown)


fmigneault:
@mishaschwartz 
Why not use the `MagpieAuth` object directly? 
https://pavics-weaver.readthedocs.io/en/latest/cli.html#authentication-example

Alternatively, could also use https://pavics-weaver.readthedocs.io/en/latest/autoapi/weaver/cli/index.html#weaver.cli.CookieAuthHandler

weaver/utils.py
```diff
@@ -1870,12 +1870,12 @@ def _request_call(method, url, kwargs):
 
 
 @cache_region("request")
-def _request_cached(method, url, kwargs):
-    # type: (AnyRequestMethod, str, RequestCachingKeywords) -> Response
+def _request_cached(method, url, kwargs, session):
+    # type: (AnyRequestMethod, str, RequestCachingKeywords, Optional[requests.Session]) -> Response
```
fmigneault, line 1874: Not sure if the session object is serializable for the caching decorator. 

Would also avoid the extra parameter. Everything should be in `RequestCachingKeywords`.

weaver/cli.py
```diff
@@ -328,8 +328,8 @@ class WeaverClient(object):
     monitor_interval = 5    # interval between monitor pooling job status requests
     auth = None  # type: AuthHandler
 
-    def __init__(self, url=None, auth=None):
-        # type: (Optional[str], Optional[AuthHandler]) -> None
+    def __init__(self, url=None, auth=None, session=None):
+        # type: (Optional[str], Optional[AuthHandler], Optional[Session]) -> None
```
fmigneault, line 332: See PR comment about using other `AuthHandler` alternatives. Seems to me that available ones are sufficient to handle the use case (passing Magpie cookies).

If they are still not sufficient for some reason (please justify if so), I would rather have a `SessionAuthHandler` that reuses the same `auth` parameter instead of adding a new one only for that use case.

fmigneault (changes requested):

mishaschwartz:
>  Why not use the MagpieAuth object directly?
> Alternatively, could also use ... [CookieAuthHandler]

Both of these make a request to a URL to obtain a token/cookie. The session strategy is for the use-case where you already have the session cookies. It's the same use-case that motivated this change (https://github.com/bird-house/birdhouse-deploy/pull/407)

weaver/cli.py
mishaschwartz: > Seems to me that available ones are sufficient to handle the use case

Unless I'm missing some other way of using these handlers, I think they cover a different use case (see: https://github.com/crim-ca/weaver/pull/597#issuecomment-1930745774)

> I would rather have a SessionAuthHandler

Actually that would be really easy. We could do something like:

```python
class SessionAuthHandler(AuthBase):
    def __init__(self, session=None):
        self.session = session

    def __call__(self, r):
        r.prepare_cookies(self.session.cookies)
        return r
```

weaver/utils.py
mishaschwartz: > Not sure if the session object is serializable for the caching decorator.

Local tests seem to show that it caches it ok.

> Everything should be in RequestCachingKeywords

Are you sure? These get passed directly to `session.request` in the underlying function and that doesn't have a `session` parameter. So we'd have to factor it out later on.

PR #599, opened 2024-02-07
add SessionAuthHandler

mishaschwartz:
Add `weaver.cli.SessionAuthHandler` class which uses the cookies stored in a pre-existing `requests.Session` instance to authenticate.

Alternative to #597 as discussed in https://github.com/crim-ca/weaver/pull/597#discussion_r1480537709

Closes #597

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -222,6 +222,24 @@ def test_auth_handler_bearer_explicit_token():
     assert resp.headers["Authorization"].startswith("Bearer") and resp.headers["Authorization"].endswith(token)
 
 
+def test_auth_handler_bearer_explicit_token_matches_request_token():
+    req_explicit_token = WebTestRequest({})
+    req_request_token = WebTestRequest({})
+    token = str(uuid.uuid4())
+    auth_explicit_token = BearerAuthHandler(token=token)
+    auth_request_token = BearerAuthHandler(identity=str(uuid.uuid4()))
+    with mock.patch(
+        "requests.Session.request",
+        side_effect=lambda *_, **__: mocked_auth_response("access_token", token)
+    ) as mock_request:
+        resp_explicit_token = auth_explicit_token(req_explicit_token)  # type: ignore
+        mock_request.assert_not_called()
+        resp_request_token = auth_request_token(req_request_token)  # type: ignore
+    assert "Authorization" in resp_explicit_token.headers and len(resp_explicit_token.headers["Authorization"])
+    assert "Authorization" in resp_request_token.headers and len(resp_request_token.headers["Authorization"])
+    assert resp_explicit_token.headers["Authorization"] == resp_request_token.headers["Authorization"]
+
+
 def test_auth_handler_cookie():
     req = WebTestRequest({})
     auth = CookieAuthHandler(identity=str(uuid.uuid4()))
@@ -273,6 +291,24 @@ def test_auth_handler_cookie_explicit_token_mapping_multi():
     assert "Cookie" in resp.headers and len(resp.headers["Cookie"])
     assert f"auth_example={token['auth_example']}" in resp.headers["Cookie"].split("; ")
     assert f"auth_example2={token['auth_example2']}" in resp.headers["Cookie"].split("; ")
+
+
+def test_auth_handler_cookie_explicit_token_matches_request_token():
+    req_explicit_token = WebTestRequest({})
+    req_request_token = WebTestRequest({})
+    token = str(uuid.uuid4())
+    auth_explicit_token = CookieAuthHandler(token=token)
+    auth_request_token = CookieAuthHandler(identity=str(uuid.uuid4()))
+    with mock.patch(
+        "requests.Session.request",
+        side_effect=lambda *_, **__: mocked_auth_response("access_token", token)
+    ) as mock_request:
+        resp_explicit_token = auth_explicit_token(req_explicit_token)  # type: ignore
+        mock_request.assert_not_called()
+        resp_request_token = auth_request_token(req_request_token)  # type: ignore
+    assert "Cookie" in resp_explicit_token.headers and len(resp_explicit_token.headers["Cookie"])
+    assert "Cookie" in resp_request_token.headers and len(resp_request_token.headers["Cookie"])
+    assert resp_explicit_token.headers["Cookie"] == resp_request_token.headers["Cookie"]
 
 
 def test_upload_file_not_found():
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -14,7 +14,6 @@ from urllib.parse import quote
 
 import mock
 import pytest
-import requests
 import yaml
 from webtest import TestRequest as WebTestRequest  # avoid pytest collect warning
 
@@ -26,7 +25,7 @@ from weaver.cli import (
     OperationResult,
     SubscriberAction,
     WeaverClient,
-    main as weaver_cli, SessionAuthHandler
+    main as weaver_cli
 )
 from weaver.formats import ContentEncoding, ContentType
 
@@ -212,6 +211,17 @@ def test_auth_handler_bearer():
     assert resp.headers["Authorization"].startswith("Bearer") and resp.headers["Authorization"].endswith(token)
 
 
+def test_auth_handler_bearer_explicit_token():
+    req = WebTestRequest({})
+    token = str(uuid.uuid4())
+    auth = BearerAuthHandler(token=token)
+    with mock.patch("requests.Session.request") as mock_request:
+        resp = auth(req)  # type: ignore
+        mock_request.assert_not_called()
+    assert "Authorization" in resp.headers and len(resp.headers["Authorization"])
+    assert resp.headers["Authorization"].startswith("Bearer") and resp.headers["Authorization"].endswith(token)
+
+
 def test_auth_handler_cookie():
     req = WebTestRequest({})
     auth = CookieAuthHandler(identity=str(uuid.uuid4()))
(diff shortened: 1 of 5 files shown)


weaver/cli.py
```diff
@@ -329,7 +342,7 @@ class WeaverClient(object):
     auth = None  # type: AuthHandler
 
     def __init__(self, url=None, auth=None):
-        # type: (Optional[str], Optional[AuthHandler]) -> None
+        # type: (Optional[str], Optional[AuthBase]) -> None
```
fmigneault, line 345: Why replace these? `AuthBase` only has this defined: 
```python
class AuthBase:
    def __call__(self, r: models.PreparedRequest) -> models.PreparedRequest: ...
```
This is very lacking in terms of abstract class, which is why `AuthHandler` with more metadata was defined.

Note also that `AnyRequest` is used, so `requests.models.PreparedRequest` cannot be assumed.

weaver/cli.py
```diff
@@ -319,6 +319,19 @@ def auth_header(self, token):
...
+    """
+    Adds the cookies from the session to the request.
+    """
+    def __init__(self, session):
+        self.session = session
+
+    def __call__(self, request):
+        # type: (PreparedRequest) -> PreparedRequest
+        request.prepare_cookies(self.session.cookies)
+        return request
```
fmigneault, line 332: I'm confused. This is exactly what `MagpieAuth` does, which is already supported. 
https://github.com/Ouranosinc/requests-magpie/blob/b26ed3c2cee4395c1bd745c7e102a442ba2199ae/requests_magpie.py#L77-L78
What is different (new use case?) that was not already covered? 
Is it only a matter of `MagpieAuth` providing some sort of `from_session` method? Why must the `session` be used instead of the cookie jar?

Linking back to https://github.com/crim-ca/weaver/pull/597#issuecomment-1930745774, there is no additional request done than the one originally performed when accessing the cookie for the first time since it gets cached in the class (see `self._cookies` in below link). 

[`MagpieAuth`](https://github.com/Ouranosinc/requests-magpie/blob/b26ed3c2cee4395c1bd745c7e102a442ba2199ae/requests_magpie.py#L47-L58) also tries to use a similar interface (on purpose) as [`AuthHandler`](https://github.com/crim-ca/weaver/blob/3a6cc5877950ed651daa779458d151d5adbb4330/weaver/cli.py#L185-L192) so that properties can be employed in the code. This interface is missing many of them.

mishaschwartz: > I'm confused. This is exactly what MagpieAuth does, which is already supported.

MagpieAuth signs in the user with their username and password:
https://github.com/Ouranosinc/requests-magpie/blob/0.2.0/requests_magpie.py#L66

There is no mechanism with MagpieAuth to include cookies that have already been acquired from somewhere else.

> Linking back to https://github.com/crim-ca/weaver/pull/597#issuecomment-1930745774, there is no additional request done than the one originally performed when accessing the cookie for the first time since it gets cached in the class (see self._cookies in below link).

I want a mechanism that doesn't make an original request to acquire the cookies _at all_.  The use-case is that the user already has the cookies, they don't need to get them again.

> [MagpieAuth](https://github.com/Ouranosinc/requests-magpie/blob/b26ed3c2cee4395c1bd745c7e102a442ba2199ae/requests_magpie.py#L47-L58) also tries to use a similar interface (on purpose) as [AuthHandler](https://github.com/crim-ca/weaver/blob/3a6cc5877950ed651daa779458d151d5adbb4330/weaver/cli.py#L185-L192) so that properties can be employed in the code. This interface is missing many of them.

That's fine, I can make this subclass AuthHandler if you'd like but it doesn't share any functionality with AuthHandler really.

weaver/cli.py
mishaschwartz: > Why replace these? AuthBase only has this defined

Is the issue that `AuthBase` is an abstract class or would you prefer that I make `SessionAuthHandler` a subclass of `AuthHandler`?

> Note also that AnyRequest is used, so requests.models.PreparedRequest cannot be assumed.

Is there a reason why it's acceptable for the MagpieAuth class but not here? https://github.com/Ouranosinc/requests-magpie/blob/0.2.0/requests_magpie.py#L56

weaver/cli.py
fmigneault: > I want a mechanism that doesn't make an original request to acquire the cookies at all. The use-case is that the user already has the cookies, they don't need to get them again.

But that would be handled with `CookieAuthHandler` no?
How is the `session` object obtained otherwise? Is it built from the Cookie value? If so, it could be passed directly to avoid this step. 

Note: 
It would be fine with adding a pre-check to `RequestAuthHandler` that skips `request_auth` call when the token/cookie is predefined in the handler if this is what is causing the issue. Ideally, all `AuthHandler` classes should be callable from either Python or the CLI, and I cannot see how to easily achieve the session object propagation in the CLI.

> That's fine, I can make this subclass AuthHandler if you'd like but it doesn't share any functionality with AuthHandler really.

This is fine. It is to be future-proof if username/identity retrieval gets added later on, such as for storing the WPS-outputs under the user-workspace. The method to do it must be defined, but the parameters would at least be available.

weaver/cli.py
fmigneault: Yes, making it a subclass of `AuthHandler` is preferable. 
The only reason it is acceptable for `MagpieAuth` is because it is an external component, so Weaver cannot push (explicitly) its definitions onto it.

weaver/cli.py
mishaschwartz: > But that would be handled with CookieAuthHandler no?

No, CookieAuthHandler still makes a request to obtain the cookie. My use case is: 

>  The use-case is that the user already has the cookies, they don't need to get them again.

> How is the session object obtained otherwise? Is it built from the Cookie value? If so, it could be passed directly to avoid this step.

Yes, that's a good example, it could be built from a cookie value (like the example given in https://github.com/bird-house/birdhouse-deploy/pull/407).
How can it be passed directly? I could build the `Cookie` header and pass that as a custom header to the WeaverClient's methods but it would be a lot nicer to be able to set this once when I initialize the WeaverClient instance.

> It would be fine with adding a pre-check to RequestAuthHandler that skips request_auth call when the token/cookie is predefined in the handler if this is what is causing the issue

Ok I can work with that.

weaver/cli.py
fmigneault: No. The idea is not to push it to an external component harder to discover. I just want to make sure this component does not reimplement something already supported by other means, which is my impression since the Cookie handler was created to pass the cookie by value for a similar use case as yours.

mishaschwartz:
Ok, here's the new usage: 

CLI:
```sh
weaver processes -aC weaver.cli.CookieAuthHandler -aT 'auth=thecookievaluethattshouldbeincluded' --url https://example.com/weaver
```

Client
```python
s = requests.Session()
# add cookies to the session
client = WeaverClient("https://example.com/weaver", auth=CookieAuthHandler(token=s.cookies.get_dict()))
```

mishaschwartz:
Note that because I made the changes in the RequestAuthHandler as discussed:

> It would be fine with adding a pre-check to RequestAuthHandler that skips request_auth call when the token/cookie is predefined in the handler

it is possible to send a token to the BearerAuthHandler as well in a similar way (if you have the bearer token already)

weaver/cli.py
```diff
@@ -289,7 +307,10 @@ def response_parser(self, response):
 
     def __call__(self, request):
         # type: (AnyRequestType) -> AnyRequestType
-        auth_token = self.request_auth()
+        if self.token is None:
+            auth_token = self.token
+        else:
+            auth_token = self.request_auth()
```
fmigneault, line 313: Isn't it supposed to be the other way around, such that the request is performed if the token was not pre-filled?

weaver/cli.py
```diff
@@ -242,6 +242,15 @@ class RequestAuthHandler(AuthHandler, HTTPBasicAuth):
...
     Base class to send a request in order to retrieve an authorization token.
     """
+
+    def __init__(self, token=None, **kwargs):
+        # type: (Any, **Any) -> None
+        if token is None:
+            self.token = token
+        else:
+            self.token = self.parse_token(token)
+        AuthHandler.__init__(self, **kwargs)
```
fmigneault, line 252: Might be better to switch the order. For example in the case of HTTP Basic auth, if the token was provided, the username and password could be inferred from the value.

weaver/cli.py
```diff
@@ -304,16 +325,40 @@ class BearerAuthHandler(RequestAuthHandler):
...
 
 class CookieAuthHandler(RequestAuthHandler):
     """
-    Adds the ``Authorization`` header formed of the authentication bearer token from the underlying request.
+    Adds the ``Cookie`` header formed from the authentication bearer token from the underlying request.
     """
 
+    @staticmethod
+    def parse_token(token):
+        # type: (Union[str, Mapping]) -> str
```
fmigneault, line 349: Use `Mapping[str, str]`?

weaver/cli.py
fmigneault: If the token was retrieved from the request, we should probably call `parse_token` below, just before calling `auth_header`. Otherwise, we get a situation where the handler could have multiple representations of the token depending on whether it came predefined or from the request.

tests/test_cli.py
```diff
@@ -225,6 +236,45 @@ def test_auth_handler_cookie():
...
+def test_auth_handler_cookie_explicit_token_string():
+    req = WebTestRequest({})
+    token = str(uuid.uuid4())
+    auth = CookieAuthHandler(token=token)
+    with mock.patch("requests.Session.request") as mock_request:
+        resp = auth(req)  # type: ignore
+        mock_request.assert_not_called()
+    assert "Authorization" not in resp.headers
+    assert "Cookie" in resp.headers and len(resp.headers["Cookie"])
+    assert resp.headers["Cookie"] == token
```
fmigneault, line 248: Could be nice to have a test that does both variant (prefilled token vs retrieved one from request) to validate that either branch results in the same `Cookie` header after parsing (see other PR comments).

weaver/cli.py
mishaschwartz: > Isn't it supposed to be the other way around,

Yes... that's embarrassing. It was the other way when I wrote the tests otherwise they would have all failed. I don't know what happened. I'll fix it

weaver/cli.py
mishaschwartz: Instead of having two calls to `parse_token` (if I implement the suggestion here: https://github.com/crim-ca/weaver/pull/599#discussion_r1483732682), I could just call `parse_token` once in `__call__` for both cases (if the token is provided or if we get if from a request)

fmigneault: Yes. That's what I had in mind as well. 👍

fmigneault (dismissed):
Thanks for the new feature (waiting for the tests to run/succeed before merge).

fmigneault:
@mishaschwartz You can add yourself to the contributors list in https://github.com/crim-ca/weaver/blob/master/AUTHORS.rst if you desire.

weaver/cli.py
```diff
@@ -2899,7 +2961,6 @@ def main(*args):
     except Exception as exc:
         msg = "Operation failed due to exception."
         err = fully_qualified_name(exc)
-        result = OperationResult(False, message=msg, body={"message": msg, "cause": str(exc), "error": err})
```
fmigneault, line 2902: Was this removed by mistake? 
The CLI should report the error that occurred.

fmigneault (changes requested):

weaver/cli.py
```diff
@@ -242,6 +242,20 @@ class RequestAuthHandler(AuthHandler, HTTPBasicAuth):
...
+        AuthHandler.__init__(self, **kwargs)
+        HTTPBasicAuth.__init__(self, username=kwargs.get("username"), password=kwargs.get("password"))
+        self.token = token
+
+    # Update __init__'s signature to include keyword arguments from AuthHandler's __init__. This is required because
+    # the CLI entrypoint inspects the auth class' signature in order to figure out which arguments to pass to it.
+    _signature_AuthHandler = inspect.signature(AuthHandler.__init__)
+    __init__.__signature__ = _signature_AuthHandler.replace(
+        parameters=(*_signature_AuthHandler.parameters.values(), inspect.signature(__init__).parameters["token"])
+    )
```
fmigneault, line 257: Seems like it would be less hackish to simply add `username, password` parameters to `__init__` as in the parent class and pass them to child classes.

fmigneault (changes requested):

weaver/cli.py
mishaschwartz: So I'm willing to do that if you'd prefer. 
But I don't really see this as a hack especially because manipulating the `__signature__` is given as an example in [PEP362](https://peps.python.org/pep-0362/#visualizing-callable-objects-signature) and if we manually add the `username`, `password`, etc. then we lose the advantage of inheriting the signature from the parent class (if the parent class's signature changes it requires updating the child classes).

weaver/cli.py
mishaschwartz: yup I saw that in the linting checks. I'll add it back in

weaver/cli.py
fmigneault: I think the literal signature is more useful since it gives direct feedback to potential users of which parameters it takes. 
I'm not too worried about changes in the parent signature since IDEs can flag those issues (at least PyCharm does), and some tests will fail if the problem is important.

fmigneault (approved):

fmigneault:
@mishaschwartz FYI https://github.com/crim-ca/weaver/tree/5.1.0

fmigneault:
@mishaschwartz Due to a CI issue with https://github.com/crim-ca/weaver/tree/5.1.0, I pushed another https://github.com/crim-ca/weaver/tree/5.1.1. The docker image for it should be ready soon.

Issue #613, opened 2024-04-03
Add more metadata mapping details in documentation


fmigneault:
## Description

Many common metadata fields correspond between CWL and OGC API - Processes.
Those should be provided more explicitly with references to relevant documents (see below).

## References

- Weaver Doc Metadata mapping: https://pavics-weaver.readthedocs.io/en/latest/package.html#metadata
- OGC Best Practices Metadata mapping: https://docs.ogc.org/bp/20-089r1.html#toc31

Issue #463, opened 2022-08-24
Parse additional metadata from schema.org in CWL document to convert into process fields


fmigneault:
## Description

Many common metadata fields can be used from https://schema.org using the following (and common) definition in CWL:
```yaml
$namespaces:
  s: https://schema.org
```

Following items could be detected and converted between CWL/OGC definitions:

- https://schema.org/softwareVersion as `s:softwareVersion` for process `version`

Additional CWL↔OGC mapping:
- https://docs.ogc.org/bp/20-089r1.html#toc31
- https://docs.ogc.org/bp/20-089r1.html#toc36

## References

- [x] relates to #31 
- [ ] relates to #158
- [ ] relates to #449
- [ ] relates to #613

PR #702, opened 2024-08-29
Parsing cwl metadata

Nazim-crim:
Changes
- Add parsing of additional metadata from schema.org in CWL document to convert into process fields

- fixes https://github.com/crim-ca/weaver/issues/463
- fixes https://github.com/crim-ca/weaver/issues/613

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,7 +3,10 @@
 Changes
 *******
 
-.. **REPLACE AND/OR ADD SECTION ENTRIES ACCORDINGLY WITH APPLIED CHANGES**
+Changes:
+--------
+- Add parsing of additional metadata from schema.org in CWL document to convert into process fields (fixes `#463 <https://github.com/crim-ca/weaver/issues/463>`_) 
+- Add more metadata mapping details in documentation (fixes `#613 <https://github.com/crim-ca/weaver/issues/613>`_) 
 
 .. _changes_latest:
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,10 +3,7 @@
 Changes
 *******
 
-Changes:
---------
-- Add parsing of additional metadata from schema.org in CWL document to convert into process fields (fixes `#463 <https://github.com/crim-ca/weaver/issues/463>`_) 
-- Add more metadata mapping details in documentation (fixes `#613 <https://github.com/crim-ca/weaver/issues/613>`_) 
+.. **REPLACE AND/OR ADD SECTION ENTRIES ACCORDINGLY WITH APPLIED CHANGES**
 
 .. _changes_latest:
 
@@ -33,6 +30,9 @@ Changes:
   and a sample `crim-ca/ncml2stac <https://github.com/crim-ca/ncml2stac/tree/main#ncml-to-stac>`_ repository
   making use of it with the `Weaver` `CLI` to generate a deployed `OGC API - Processes` definition
   (fixes `#63 <https://github.com/crim-ca/weaver/issues/63>`_).
+- Add parsing of additional metadata from schema.org in CWL document to convert into process fields
+  (fixes `#463 <https://github.com/crim-ca/weaver/issues/463>`_) 
+- Add more metadata mapping details in documentation (fixes `#613 <https://github.com/crim-ca/weaver/issues/613>`_) 
 
 Fixes:
 ------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -31,8 +31,8 @@ Changes:
   making use of it with the `Weaver` `CLI` to generate a deployed `OGC API - Processes` definition
   (fixes `#63 <https://github.com/crim-ca/weaver/issues/63>`_).
 - Add parsing of additional metadata from schema.org in CWL document to convert into process fields
-  (fixes `#463 <https://github.com/crim-ca/weaver/issues/463>`_) 
-- Add more metadata mapping details in documentation (fixes `#613 <https://github.com/crim-ca/weaver/issues/613>`_) 
+  (fixes `#463 <https://github.com/crim-ca/weaver/issues/463>`_).
+- Add more metadata mapping details in documentation (fixes `#613 <https://github.com/crim-ca/weaver/issues/613>`_).
 
 Fixes:
 ------
--- a/docs/source/package.rst
+++ b/docs/source/package.rst
@@ -1307,25 +1307,20 @@ Below is a list of compatible elements.
 +-----------------------------------------+----------------------------------------------------------+
 | ``metadata``                            | ``$schemas``/``$namespace``                              |
 | (using ``title`` and ``href`` fields)   | (using namespace name and HTTP references)               |
+|                                         | ``s:author``                                             |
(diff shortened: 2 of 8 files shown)


tests/processes/test_wps_package.py
```diff
@@ -667,3 +668,84 @@ def test_mask_process_inputs(inputs, expect):
...
+            "cwl_package_package": {
+                "s:codeRepository": "https://gitlab.com/",
+            },
+            "wps_package_metadata": {}
+        },
+        {
+            "metadata": [
+                {
+                    "rel": "codeRepository",
+                    "href": "https://gitlab.com/"
```
fmigneault, line 710: For this case, I think it is safe to assume `type: text/html` could be used. You can set it by default for `codeRepository`. 

Consider adding a `s:license` test case as well. In that case, you can use
`weaver.format.get_content_type(os.path.splitext(<URL>)[-1], default=ContentType.TEXT_PLAIN)` to resolve the specified `type`.
A `https://example.com/LICENSE.md` should result in `text/markdown`.

tests/processes/test_wps_package.py
```diff
@@ -667,3 +668,84 @@ def test_mask_process_inputs(inputs, expect):
...
+])
+def test_process_metadata(original, expected):
+    # type: (CWL, CWL) -> None
+    cwl_package_package = original["cwl_package_package"]
+    wps_package_metadata = original["wps_package_metadata"]
+    _update_package_metadata(wps_package_metadata, cwl_package_package)
+    # Assertions
+    for key in expected:
+        assert key in wps_package_metadata
+        assert sorted(wps_package_metadata[key]) == sorted(expected[key])
```
fmigneault, line 751: `assert wps_package_metadata == expected` directly does not work?

weaver/processes/wps_package.py
```diff
@@ -708,6 +718,37 @@ def _update_package_metadata(wps_package_metadata, cwl_package_package):
             set(wps_package_metadata.get("keywords", [])) | set(cwl_package_package.get("s:keywords", []))
         )
 
+    # specific use case with a different mapping
+    if "s:version" in cwl_package_package:
+        wps_package_metadata["version"] = str(
+             str(cwl_package_package.get("s:version", "")) or str(wps_package_metadata.get("version"))
```
fmigneault, line 724: I think those must be switched around. The process `version` explicitly provided takes priority over the CWL one.

weaver/processes/wps_package.py
```diff
@@ -708,6 +718,37 @@ def _update_package_metadata(wps_package_metadata, cwl_package_package):
...
+        )
+
+    for metadata_mapping in SUPPORTED_METADATA_MAPPING:
+        if metadata_mapping in cwl_package_package:
+            metadata = wps_package_metadata.get("metadata", [])
+            if (isinstance((cwl_package_package[metadata_mapping]), str)
+                    and urlparse(cwl_package_package[metadata_mapping]) != ""):
```
fmigneault, line 731: For consistency with the rest of the code, use : 

```python
if (
    isinstance((cwl_package_package[metadata_mapping]), str)
    and urlparse(cwl_package_package[metadata_mapping]) != ""
):
    [...]
```

weaver/typedefs.py
```diff
@@ -220,6 +220,14 @@ class CWL_SchemaName(Protocol):
         "$namespaces": NotRequired[CWL_Namespace],
         "$schemas": NotRequired[Dict[str, str]],
         "$graph": NotRequired[CWL_Graph],
+        "s:author": NotRequired[List[Dict[str, JSON]]],
+        "s:citation": NotRequired[List[Dict[str, str]]],
```
fmigneault, line 224: Looks like it should be `NotRequired[List[Union[str, JSON]]` based on https://schema.org/citation.

weaver/processes/wps_package.py
fmigneault: And, I think it should be `urlparse(cwl_package_package[metadata_mapping]).scheme`, since `urlparse` returns an object, therefore always `!= ""`. 

Try for example with `{"s:dateCreated": "2024-08-29"}`. It should enter this `if`, when the `else` is expected.

fmigneault (changes requested):
Need to add an entry to `CHANGES.md`.

tests/processes/test_wps_package.py
Nazim-crim: So the type should be added to the metadata when it's a link ? And if its a codeRepository we can always assume the html type without using `weaver.format.get_content_type(os.path.splitext(<URL>)[-1], default=ContentType.TEXT_PLAIN)`
```
{
    "type": "type: text/html"
    "rel": "codeRepository",
    "href": "https://gitlab.com/"
}
```

fmigneault: `"type": "text/html"`
And yes, it can be assumed to simplify things.

CHANGES.rst
```diff
@@ -30,6 +30,9 @@ Changes:
   and a sample `crim-ca/ncml2stac <https://github.com/crim-ca/ncml2stac/tree/main#ncml-to-stac>`_ repository
   making use of it with the `Weaver` `CLI` to generate a deployed `OGC API - Processes` definition
   (fixes `#63 <https://github.com/crim-ca/weaver/issues/63>`_).
+- Add parsing of additional metadata from schema.org in CWL document to convert into process fields
+  (fixes `#463 <https://github.com/crim-ca/weaver/issues/463>`_) 
+- Add more metadata mapping details in documentation (fixes `#613 <https://github.com/crim-ca/weaver/issues/613>`_)
```
fmigneault, line 35: Add the leading `.` instead of the space, for each item.

docs/source/package.rst
```diff
@@ -1312,6 +1312,20 @@ Below is a list of compatible elements.
...
+
+Parameters in :term:`CWL` Context:
+
+- ``s:author``
+- ``s:citation``
+- ``s:codeRepository``
+- ``s:contributor``
+- ``s:dateCreated``
+- ``s:license``
+- ``s:releaseNotes``
```
fmigneault, line 1328: For this, I believe it should be added directly to the ``metadata`` line that is already present in the table. Basically, just extend the right side (where ``$schemas``/``$namespace`` is indicated) with the mention of those fields. 

You can also add the `[#cwl_schemaorg]_` footnote reference (as in the ``s:keywords`` case), that points to the CWL metadata example below, which contains all of those fields.

docs/source/package.rst
```diff
@@ -1312,6 +1312,20 @@ Below is a list of compatible elements.
 +-----------------------------------------+----------------------------------------------------------+
 | ``abstract``/``description``            | ``doc``                                                  |
 +-----------------------------------------+----------------------------------------------------------+
+| ``version``                             | ``s:version``/``s:softwareVersion``                      |
```
fmigneault, line 1315: Could also use the `[#cwl_schemaorg]_`.

fmigneault (commented):
Just editorial edits left.

fmigneault (approved):

fmigneault:
@Nazim-crim Thanks!

Issue #412, opened 2022-03-14
Support OGC-API Processes - Part 3: Workflows


fmigneault:
## Description

Posting a workflow as a *Nested Process* definition of processes could be supported.
This would align with Part 3 ad-hoc Workflows, but used for immediate execution rather than deployment.

```json
{
  "process": "https://server.com/processes/ProcessID",
  "inputs": {
    "in": {
      "process": "https://server.com/processes/OtherNested",
      "inputs": {
        "data": 123
      }
    }
  }
}
``` 

This feature addresses only the *Nested Processes* (and derived requirements). The other Part 3 features such as Input/Output Collections are tracked by : 
- #682 
- #683

ToDo: 
- [ ] update the `ogcapi_proc_part3` conformance items accordingly.
- [x] #747
- [ ] #750
- [ ] #640
- [x] support *Remote Processes* requirement (de facto in our case)

## References

- http://www.opengis.net/spec/ogcapi-processes-3/1.0/req/workflows
- https://github.com/opengeospatial/ogcapi-processes/tree/master/extensions/workflows
- https://github.com/opengeospatial/ogcapi-processes/tree/master/extensions/workflows/requirements/nested-processes
- https://docs.ogc.org/DRAFTS/21-009.html#_requirement_class_nested_processes

PR #748, opened 2024-10-30
Job nested process workflow

### LINKED ISSUE #747: support the Nested Processes as `process: <uri>` construct

fmigneault:
- fixes #747
- relates to #412
- depends on #742 
- fixes [GD-567](https://crim-ca.atlassian.net/browse/GD-567)
- fixes [OGC09-18](https://crim-ca.atlassian.net/browse/OGC09-18)


weaver/processes/ogc_api_process.py
```diff
@@ -1,24 +1,65 @@
...
+        # the 'type' is expected by the output staging operation to retrieve a file as remote URI
+        # any other 'type' does nothing, but a valid CWL type is needed to avoid raising the validation
+        for output in expected_outputs.values():
+            cwl_type = PACKAGE_FILE_TYPE if "format" in output else "string"
+            output.setdefault("type", cwl_type)
+
+    def format_inputs(self, job_inputs):
+        # type: (JobInputs) -> ExecutionInputsMap
+        inputs = convert_input_values_schema(job_inputs, JobInputsOutputsSchema.OGC)
+        return inputs
```
Nazim-crim, line 52: Why use the same `JobInputsOutputsSchema ` for input and output while each have different converting function `convert_input_values_schema` and `convert_output_params_schema`

weaver/processes/wps_process_base.py
```diff
@@ -465,19 +471,41 @@ def dispatch(self, process_inputs, process_outputs):
             "response": ExecuteResponse.DOCUMENT,
             "inputs": process_inputs,
         }
+        execute_headers = {
+            "Prefer": "respond-async",
+            "Accept": ContentType.APP_JSON,
+            "Content-Type": ContentType.APP_JSON,
+        }
```
Nazim-crim, line 478: This header will never be updated ?

weaver/wps_restapi/swagger_definitions.py
```diff
@@ -3935,6 +3936,70 @@ class ExecuteCollectionInput(FilterSchema, SortBySchema, PermissiveMappingSchema
...
+    _schema_extra = {
+        "type": null,
+        "title": "ExecuteNestedProcessParameters",
+        "$ref": f"{OAS3DefinitionHandler.json_pointer}ExecuteProcessParameters"
+    }
+
+    @colander.deferred
+    def _children(self, __bindings):
+        # type: (Dict[str, Any]) -> List[colander.SchemaNode]
+        self.children = [node.clone() for node in ExecuteProcessParameters().children]
```
Nazim-crim, line 3966: Isn't confusing calling it a bind if it's not a shallow copy ?
Shouldn't it be` self.children = list(ExecuteProcessParameters().children)`

Nazim-crim (approved):
lgtm

weaver/processes/ogc_api_process.py
fmigneault: The schema represents the use of list vs mapping representation. This is identical on both side. However, the "inputs" in this case are the submitted data, while outputs are the desired value/ref/format definitions. The 2 functions handle this content differently and separately.

weaver/processes/wps_process_base.py
fmigneault: Not in this case because the pipeline always expect to receive the "document" response to extract all the results available from the job in the same matter, whether there is only one or many.

weaver/wps_restapi/swagger_definitions.py
fmigneault: Limitation from the parent `colander` library that called the method `bind`. Nested nodes have to be cloned, otherwise you could have by-ref conflicting updates if similar definitions are used simultaneously.

fmigneault: I've added a test in [`ef93c74` (#748)](https://github.com/crim-ca/weaver/pull/748/commits/ef93c7406bf3ef18a997a642824f5c14fdc6bc65) so you can get a better understanding of the feature and its behavior.

PR #789, opened 2025-01-24
add Job HTML response

### LINKED ISSUE #779: HTML Job status response
## Description

The process listing, description, and job listing all have HTML responses support. However, when listing jobs (`/jobs?f=html` or by default when in a browser), the provided links of the listed `JobID` point to `/jobs/{jobID}?f=html`. Since HTML is not implemented for that endpoint, Weaver's own references result in 406, and users are forced to manually set `f=json` to obtain the status contents as JSON.

## To Do

- [ ] Implement the Job Status response as HTML, such that the UI navigation in browsers is more natural. 
- [ ] Given that jobs contain *many* other sub-paths, not all easily represented by HTML, they should ensure to omit the `f=html` (or maybe even explicitly set `f=json`?).

## References

- relates to https://crim-ca.atlassian.net/browse/GD-571

fmigneault:
- fixes #779

### DIFF
--- a/weaver/wps_restapi/jobs/jobs.py
+++ b/weaver/wps_restapi/jobs/jobs.py
@@ -317,6 +317,16 @@ def trigger_job_execution(request):
     return submit_job_dispatch_task(job, container=request, force_submit=True)
 
 
+@sd.provider_jobs_service.get(
+    tags=[sd.TAG_JOBS, sd.TAG_STATUS, sd.TAG_PROVIDERS],
+    schema=sd.GetProviderJobEndpoint(),
+    accept=ContentType.TEXT_HTML,
+    renderer="weaver.wps_restapi:templates/responses/job_status.mako",
+    response_schemas=sd.derive_responses(
+        sd.get_provider_single_job_status_responses,
+        sd.GenericHTMLResponse(name="HTMLProviderJobStatus", description="Job status.")
+    ),
+)
 @sd.provider_job_service.get(
     tags=[sd.TAG_JOBS, sd.TAG_STATUS, sd.TAG_PROVIDERS],
     schema=sd.GetProviderJobEndpoint(),
@@ -328,7 +338,17 @@ def trigger_job_execution(request):
     response_schemas=sd.get_provider_single_job_status_responses,
 )
 @sd.process_job_service.get(
-    tags=[sd.TAG_PROCESSES, sd.TAG_JOBS, sd.TAG_STATUS],
+    tags=[sd.TAG_JOBS, sd.TAG_STATUS, sd.TAG_PROCESSES],
+    schema=sd.GetProcessJobEndpoint(),
+    accept=ContentType.TEXT_HTML,
+    renderer="weaver.wps_restapi:templates/responses/job_status.mako",
+    response_schemas=sd.derive_responses(
+        sd.get_single_job_status_responses,
+        sd.GenericHTMLResponse(name="HTMLProcessJobStatus", description="Job status.")
+    ),
+)
+@sd.process_job_service.get(
+    tags=[sd.TAG_JOBS, sd.TAG_STATUS, sd.TAG_PROCESSES],
     schema=sd.GetProcessJobEndpoint(),
     accept=[ContentType.APP_JSON] + [
         f"{ContentType.APP_JSON}; profile={profile}"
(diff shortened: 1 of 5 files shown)


requirements-dev.txt
```diff
@@ -39,6 +41,7 @@ pylint>=3,<4
 pylint_quotes @ git+https://github.com/marekhanus/pylint-quotes.git@0.3.0a2
 pylint-per-file-ignores
 responses
+# FIXME: pin >3.3.1 when available to remove 'marshmallow' pin (https://github.com/pyupio/safety/issues/711)
```
languith, line 44: I'm not sure which line this comment targets. Can you clarify or delete it if it repeats of line 19?

weaver/wps_restapi/jobs/jobs.py
```diff
@@ -348,16 +379,34 @@ def trigger_job_execution(request):
...
+    if schema == JobStatusSchema.OPENEO:
+        cwl_url = get_path_kvp(f"{job.process_url(request)}/package", f=OutputFormat.JSON)
+        job_prop = {"process": {"title": "CWL Application Package", "href": cwl_url, "type": ContentType.APP_CWL_JSON}}
+    job_body = job.json(request, **job_prop)
     if schema == JobStatusSchema.OPENEO:
+        # additional properties that are not validated explicitly
+        # align the content with metadata schema
+        # status is defined here to limit the combinations reported in OpenAPI as OGC-only statuses
+        job_body["$schema"] = sd.OPENEO_API_SCHEMA_JOB_STATUS_URL
+        job_body["type"] = JobStatusSchema.OPENEO
```
languith, line 402: This function is hard to read due to the `if` repetition. 
Does something like this work better?
```python
job = get_job(request)
schema, headers = get_job_status_schema(request)
if schema == JobStatusSchema.OPENEO:
    cwl_url = get_path_kvp(f"{job.process_url(request)}/package", f=OutputFormat.JSON)
    job_body = job.json(
        request,
        process={"title": "CWL Application Package", "href": cwl_url, "type": ContentType.APP_CWL_JSON}
    )
    job_body["$schema"] = sd.OPENEO_API_SCHEMA_JOB_STATUS_URL
    job_body["type"] = JobStatusSchema.OPENEO
    job_body["status"] = map_status(job_body["status"], StatusCompliant.OPENEO)
else:
    job_body = job.json(request)

return HTTPOk(json=job_body, headers=headers)
```

requirements-dev.txt
fmigneault: It's for the `safety` package under it. Temporary workaround needed within `safety`'s `marshmallow` dependency. 

The required `safety>3.3.1` has been released since, so I can try to fix it. 
Will do in a separate PR, and this would be removed.

weaver/wps_restapi/jobs/jobs.py
fmigneault: Yes, that works, but I will add comments since the subtlety of validation happening under `job.json` is what prompted these subsequent `if`.

requirements-dev.txt
fmigneault: will use upstream changes from https://github.com/crim-ca/weaver/pull/814

Nazim-crim (dismissed):

weaver/datatype.py
```diff
@@ -1659,6 +1668,7 @@ def json(self, container=None):  # pylint: disable=W0221,arguments-differ
             "progress": int(self.progress),
             "links": self.links(settings, self_link="status")
         }
+        job_json.update(**kwargs)
```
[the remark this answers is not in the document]
fmigneault: This is to allow overriding some definitions.
If we did `job.status = "finished"` followed by `job.json()`, the above `map_status` would switch it back to the OGC-status mapping using `succeeded` instead.

Another reason is if we want to provide extended properties in the response, with something that is allowed by the `JobStatusInfo` schema, but is not explicitly generated by the above `job.json()` properties.

weaver/wps_restapi/swagger_definitions.py
```diff
@@ -289,6 +290,30 @@
 ]
 PROVIDER_DESCRIPTION_FIELD_AFTER = ["links"]
 
+JOB_STATUS_FIELD_FIRST = ["jobID", "processID", "providerID"]
+JOB_STATUS_FIELD_AFTER = [
```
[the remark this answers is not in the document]
fmigneault: `success` is a utility property of the class, but is not rendered in the JSON response. This list is only for ordering the properties returned.

PR #877, opened 2026-02-10
pin `setuptools` to resolve its `pkg_resources` dependency still required by `pyramid`

fmigneault:
- relates to https://github.com/Pylons/pyramid/issues/3731
- relates to https://github.com/pypa/setuptools/pull/5007


fmigneault:
Ignore `check-links` error, to be addressed in separate PR.

PR #897, opened 2026-03-02
Fix get authorized file for multi token

### LINKED ISSUE #894: X-Auth-Vault multi_token handling
## Describe the bug

When using the `X-Auth-Vault` header generated by the Weaver client’s `_upload_files` method, the header cannot be used to retrieve the file, resulting in an `HTTP 403 Forbidden error`.

Accessing a file using a multi-token header on GET /vault/<file_id>:

```
curl -X 'GET' \
  'http://instance_ip:4001/vault/11832a5f-29ad-45f3-a871-05dc5773f3e2' \
  -H 'accept: application/json' \
  -H 'X-Auth-Vault: token bae25bae314d052eb3b5899f199079e905cee6db197f99ebd95e08bef8a77724; id=73e7ce13-127d-4e39-a034-2fc769b3164a,token 3d539789889e2be3e230fb750010be44a7d1f1c7f934664ca06e27f46ff50200; id=11832a5f-29ad-45f3-a871-05dc5773f3e2'
```
Error : 

Image

When accessing the file using a single token, the file can be accessed successfully:

```
curl -X 'GET' \
  'http://<instance_ip>:4001/vault/11832a5f-29ad-45f3-a871-05dc5773f3e2' \
  -H 'accept: application/json' \
  -H 'X-Auth-Vault: token 3d539789889e2be3e230fb750010be44a7d1f1c7f934664ca06e27f46ff50200'
```
Returns: 200 OK

## How to Reproduce
Execute a process that uploads two or more files to the vault.
Try to access one of the files using a header containing multiple tokens.
The request will fail with a 403 Forbidden error.

## Expected behavior

Accessing a file in the vault using multi token in the X-Auth-Vault should be handled. 

## Context

- Instance: Weaver API
- Version : 6.8.3

Nazim-crim:
Resolves : https://github.com/crim-ca/weaver/issues/894

- Fix multi-token Vault authentication header parsing to support both single and multiple file access tokens.
  The ``parse_vault_token`` function now handles plain token strings (e.g., from WPS process context) and full
  header formats (e.g., ``token <value>; id=<uuid>``), and correctly validates token presence for the requested
  file UUID. The mismatch detection logic was updated to properly check if the requested file ID exists in the
  parsed tokens rather than only validating against the first token key

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,7 +16,12 @@ Changes:
 
 Fixes:
 ------
-- No change.
+- Fix multi-token Vault authentication header parsing to support both single and multiple file access tokens.
+  The ``parse_vault_token`` function now handles plain token strings (e.g., from WPS process context) and full
+  header formats (e.g., ``token <value>; id=<uuid>``), and correctly validates token presence for the requested
+  file UUID. The mismatch detection logic was updated to properly check if the requested file ID exists in the
+  parsed tokens rather than only validating against the first token key
+  (fixes `#897 <https://github.com/crim-ca/weaver/issues/897>`_).
 
 .. _changes_6.9.0:
 
--- a/tests/test_vault.py
+++ b/tests/test_vault.py
@@ -71,6 +71,63 @@ def test_parse_vault_token(header, unique, expected):
 
 
 @pytest.mark.vault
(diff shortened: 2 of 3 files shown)


fmigneault (changes requested):

fmigneault (approved):
@Nazim-crim Thanks!

PR #912, opened 2026-04-21
Support /results/{id}/{N}

### LINKED ISSUE #759: Support `/results/{id}/{N}` and `/outputs/{id}/{N}` endpoints
## Description

Given an output that happens to be an "array" (or rather, a multi-value result of any data format or media-type representation), adding an index sub-path could allow retrieving that specific reference. This could be useful for extracting a specific result used for chaining to another (or multiple) process(es).

In other words: 

`/jobs/{jobId}/results`
```json
{
  "result": [1, 2, 3]
  "urls": ["https://.../1", "https://.../2"]
}
```
`/jobs/{jobId}/results/result`
```json
[1, 2, 3]
```
`/jobs/{jobId}/results/result/1`
```json
2
```
`/jobs/{jobId}/results/urls`
```json
["https://.../1", "https://.../2"]
```
`/jobs/{jobId}/results/result/0`
```yaml
https://.../1
```

## To Do

- [ ] implement the endpoint by index
- [ ] raise an "out-of-range" error if the index is invalid or does not apply (not an array value)
- [ ] handle content-negotiation as applicable for that indexed-value

## References

- relates to https://github.com/opengeospatial/ogcapi-processes/issues/414
- depends on https://github.com/crim-ca/weaver/issues/18 (prefix endpoints by ID)

Nazim-crim:
- Add support for indexed array access to job results via ``/results/{output_id}/{index}`` endpoints
  - fixes https://github.com/crim-ca/weaver/issues/759

  This allows retrieving individual elements from array-type job outputs using zero-based indexing.
  Available on job-only, process-scoped, and provider-scoped result endpoints:

  - ``/jobs/{jobID}/results/{output_id}/{index}``
  - ``/processes/{processID}/jobs/{jobID}/results/{output_id}/{index}``
  - ``/providers/{providerID}/processes/{processID}/jobs/{jobID}/results/{output_id}/{index}``

  The endpoint returns HTTP 400 for invalid or out-of-range indices, HTTP 404 when the output is not found,
  and HTTP 422 when attempting to index into non-array outputs.

  Added utility function ``get_job_result_by_index()`` in ``weaver.wps_restapi.jobs.utils`` to handle
  indexed result retrieval with comprehensive validation and error handling.


weaver/wps_restapi/jobs/utils.py
```diff
@@ -862,6 +863,94 @@ def get_job_results_response(
...
+            "type": "http://www.opengis.net/def/exceptions/ogcapi-processes-1/1.0/invalid-parameter",
+            "detail": f"Index {index} is out of range for output '{output_id}' (length: {len(output_value)}).",
+            "status": HTTPBadRequest.code,
+            "cause": {"index": index, "length": len(output_value), "output": output_id}
+        })
+
+    # Extract the indexed element from the array
+    indexed_element = output_value[index]
+
+    # Return the indexed element as JSON
```
fmigneault: remove superflus comments, same for others above since their respective errors are already descriptive enough about what they check

weaver/wps_restapi/jobs/utils.py
```diff
@@ -862,6 +863,94 @@ def get_job_results_response(
...
+    :raises HTTPNotFound: If the output ID is not found in the job results.
+    :raises HTTPUnprocessableEntity: If the output is not an array.
+    """
+    raise_job_dismissed(job, container)
+    raise_job_bad_status_success(job, container)
+
+    # Validate index is non-negative
+    if index < 0:
+        raise HTTPBadRequest(json={
+            "title": "InvalidIndex",
```
fmigneault: For each of those `title`, I think we should add a "Job Output" or similar before it so it is very explicit about which index or other problem we are talking about. They can be more "human readable", like "Job Output Invalid Index" since the `type` already provides an "error code".

tests/wps_restapi/test_jobs.py
```diff
@@ -1932,6 +1932,313 @@ def test_job_outputs_response(self):
         assert resp.status_code == 200
         assert resp.json["outputs"] == {"test": {"value": "data"}}
 
+    @pytest.mark.oap_part4
```
fmigneault: Part 1 for all of them.
They could also have the `job` marker.

fmigneault (approved):

Issue #993, opened 2026-07-06
Support `profile` links, inputs and outputs


fmigneault:
## Description

Whenever a link `{"href":"...", "type":"..."}` is provided a `profile` may be supplied to expland it further.

Some places to consider:
- [ ] process description link within process listing, job status or job results
- [ ] process description itself when offering `rel=alternate` 
- [ ] specific job outputs (`/jobs/{id}/results/{outputID}[/{N}]`) if they have another known profile representation

## References

- relates to https://github.com/opengeospatial/ogcapi-processes/issues/550
- relates to #984

Issue #990, opened 2026-06-29
Support multipart per-part process metadata


fmigneault:
## Description

With the multipart implementation (#968), one can deploy a CWL Workflow, its sub-CWL tools, and an (*potentially*) optional Process Description that describes it.

For the time being, additional process metadata are parsed/validated, but not forwarded in the deployment along the CWL (as it does in the case of a typical `processDescription`/`executionUnit` deploy combination).

Also, in Weaver, each sub-CWL tool is itself a distinct process deployment. Therefore, each of those parts could themselves support an accompanying Process Description.

This feature consists of supporting that capabilty, by cross-linking the `id` shared between corresponding CWLs and the `id` in the Process Description. If the `id` is not provided in the Process Description, it is safe to assume for backward compatibility that it was meant for the main workflow. If the `id` is not provided in the CWL definition, it may be resolved from `Content-ID` or `Content-Location` (this resolution portion should already be implemented). 

## To Do

- [ ] Support parts indicating `Content-Profile: https://www.opengis.net/def/profile/OGC/0/ogc-process-description` 
  (see `sd.OGC_API_PROC_PROFILE_PROC_DESC_URI`)
- [ ] Ensure that the `Content-Type: application/json` is supported.
- [ ] If `Content-Type: application/yaml` is indicated, the `OutputFormat.convert` tool can be used to transform inline.
- [ ] Each part asserting the process description profile must validate against `sd.ProcessDeployment` schema
- [ ] Process descriptions (main or sub part) must be matched against corresponding CWL parts for their per-process specific deployment
- [ ] Resulting process deployment must have the relevant details combined (already implemented during deployment, but check it works for sub-parts resolved through multipart).
- [ ] Ensure that error messages of #997 are aligned with any additional process description metadata (notably, for disambiguation between `application/json` of Execute vs ProcessDescription)
- [ ] Support deployment of (`processDescription`/`executionUnit`) as single JSON part when provided with the relevant profile negotiation. 
  - Such part would be semantically equivalent to 2 distinct CWL/ProcessDescription parts of matching `id`. 
  - WIP: see `Content-Profile: https://www.opengis.net/def/profile/OGC/0/ogc-deploy-process`
    (https://github.com/opengeospatial/ogcapi-processes/issues/616)

## References

- base implementation #968
- relates to https://github.com/crim-ca/weaver/issues/977

Issue #984, opened 2026-06-20
Support KVP `?response[profile]` parameter


fmigneault:
## Description

- [ ] Support KVP `?response[profile]` parameter during process execution submission.
- [ ] Support the parameter (+ `?profile` alias) on the Job Results endpoint directly.
- [ ] Alias `?profile` is NOT used on execution endpoint (since it can be confused with the *input* content/execution profile)
- [ ] The `Accept-Profile` is used as fallback if provided, but the query parameter takes precedence.
- [ ] The Job/Results response format is adjusted based on the requested profile.
- [ ] The response includes the corresponding `Link: rel=profile` and similar alternate definitions (`Content-Profile`, etc.)

## References

- relates to https://github.com/crim-ca/weaver/issues/445
- relates to https://github.com/opengeospatial/ogcapi-processes/issues/550
- relates to https://github.com/opengeospatial/ogcapi-processes/issues/557

PR #997, opened 2026-07-14
Support ad-hoc CWL workflows  

### LINKED ISSUE #834: Support ad-hoc CWL workflows
## Description

Job submission of an ad-hoc workflow is supported in Weaver: https://github.com/crim-ca/weaver/issues/716. However, this expects the *OGC API - Processes* Nested Processes format (https://github.com/crim-ca/weaver/issues/412, https://github.com/crim-ca/weaver/issues/747), and doesn't allow directly using CWL.

In order to allow users to interact with Weaver consistently using CWL, whether they design a single process, a deployed workflow, or submitting a new workflow, Weaver would need to handle *ad-hoc Workflows as CWL* (i.e.: [OGC API - Processes - Part 3: Workflows and Chaining / Requirements classes specifying alternative ad-hoc workflow definitions](https://docs.ogc.org/DRAFTS/21-009.html#_766885ff-ad3f-2111-5f7c-78c66514a2e4)). Although OGC defines the ad-hoc CWL requirement, that does not make much sense at the moment (see below).

---

Parsing a CWL submitted to `POST /processes` to deploy it is fairly simple, given that CWL can already be understood by Weaver. However, the submission of the *input values* for that Workflow on `POST /jobs` is not really defined, given that CWL *Workflow Definition* and the *Job Parameters* are separate documents. Further exploration is needed to establish how to provide *both simultaneously and directly* in the request contents of `POST /jobs`.

In other words, what would be the preferred strategy to send to `POST /jobs` both the CWL and Execute contents: 

```json
{
  "doc": "The Workflow *Definition*",
  "class": "Workflow", 
  "inputs": { "<input-id>": {"type": "<cwl-type>"} },
  "outputs": { "<output-id>": {"type": "<cwl-type>"} },
  "steps": { "step-1": "{...}", "step-2": "{...}" }
}
```
```json
{ 
  "description": "The Workflow *Values* / Job Parameters",
  "inputs": { "<input-id>": "the-value" },
  "outputs": { "<output-id>": { "format": { "mediaType": "iana:application/x-what-i-want" } } }
}
```

---

Furthermore, using an ad-hoc CWL could limit the use of OGC-specific inputs, such as `collection` URLs (https://github.com/crim-ca/weaver/issues/682) that could not be *directly* submitted in a CWL that has no handling of such concept (`type: File/Directory` only).

---

Finally, the ad-hoc Job submission assumes (as of time of writing) that a `process` URL can be resolved at the root of the request's body, to know which Workflow or Atomic __*pre-deployed*__ process is to be called. With an ad-hoc CWL that defines a graph (rather than the tree-like nested process approach, https://github.com/opengeospatial/ogcapi-processes/issues/421), a workflow would be "*virtually and dynamically*" created. How this Workflow gets referenced for *Provenance* aspects (https://github.com/crim-ca/weaver/issues/673), and how to reference it where `processID` would currently be expected, is not established.

A possible approach would be to actually deploy the ad-hoc Workflow behind the scene. The resulting process could be attributed `processID = job:<jobID>` (leveraging #837 feature). Then, every other feature would resolve as normal, referencing an actual process. The use of `job:` prefix could allow doing some internal management, such as cleaning up (undeploy) this ad-hoc process/workflow on `DELETE /jobs/<jobID>` and omitting it by default from `GET /processes` listing (except if `?jobs=true` is passed) to avoid polluting the response with dynamic processes.

---

## Potential Approaches

1. Use `multipart/related` (or similar) to submit both the `application/cwl(+json|+yaml)` and the relevant `application/json` (https://github.com/opengeospatial/ogcapi-processes/blob/master/openapi/schemas/processes-workflows/execute-workflows.yaml) with their respective schemas.
   - Most realistically the safest (and easiest to interpret?) approach, given that each concept are kept separate. 
   - Involves a slightly more complex request definition by clients, since multipart/boundaries nomenclature is involved rather than a plain JSON, though libraries are typically available to handle these aspects
   - See https://github.com/opengeospatial/ogcapi-processes/pull/588
2. Using some mixed representation, such as nesting the `value`/`href`/`collection` within the CWL
   - possible attribute conflict? 
     (e.g.: CWL's `format` vs Execute's `format` for outputs? - maybe not since no `format` for CWL _**outputs**_)
   - even if not conflicting, potential confusion about which parameter comes from which spec?
   - needs to be parsed/removed before loading in `cwltool` (avoid unrecognized parameters by CWL)
   - doesn't allow the direct-value of OGC (i.e.: `"<input-id>": 1234`)
    (values must be nested under `value` parameter explicitly since other CWL parameters would be present under the  `"<input-id>": {...}` object)
3. Using the `valueFrom` CWL property, to directly inject the values
   - Given `collection` input, the specific `valueFrom` value might be ambiguous or malformed in CWL-proper schemas. They would also need to be extracted before validation
   - Might cause ambiguities with CWL that actually employ `valueFrom` (how to distinguish them from the others used only for ad-hoc submission?)
   - Most probably does not allow the full extent of additional parameters that could be supplied along the OGC input values (`contentMediaType`, `contentSchema`, etc.)
4. ❓ *open to suggestions*

## References

The following references are potential candidates or related concerns that could impact the result of this feature.

- https://github.com/crim-ca/weaver/issues/754 
- https://github.com/opengeospatial/ogcapi-processes/issues/500
- https://github.com/opengeospatial/ogcapi-processes/issues/445
- https://github.com/opengeospatial/ogcapi-processes/issues/304 
- https://github.com/opengeospatial/ogcapi-processes/issues/279
- https://github.com/opengeospatial/ogcapi-processes/issues/421

## Selected Approach

- https://github.com/opengeospatial/ogcapi-processes/issues/604

Nazim-crim:
https://crim-ca.atlassian.net/browse/NRCAN06-88

- Add support for ad-hoc `CWL` workflow execution through the ``POST /jobs`` endpoint using ``multipart/mixed``
  or ``multipart/related`` content types. Users can now deploy and execute a `Process` in a single request by
  submitting both the `CWL` workflow definition (with ``Content-Profile: http://www.opengis.net/def/ogcapi-processes/2.0/process-description``)
  and execution parameters (with ``Content-Profile: http://www.opengis.net/def/ogcapi-processes/2.0/execute``)
  in a multipart request body. The workflow is automatically deployed (with temporary ``ad-hoc`` tagging) before
  execution, eliminating the need for separate deployment and execution steps for one-time workflow runs
 
- resolves #834


docs/source/package.rst
```diff
@@ -681,6 +681,123 @@ Example multipart request structure:
     - :ref:`proc_ogc_api_multi_cwl` for deployment request examples
     - |ogc-api-proc-part2|_ CWL Multipart Content conformance class
 
+.. _app_pkg_adhoc:
+
+Ad-hoc Workflow Execution
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
fmigneault: Can you move the section under `processes.rst`.
I think it would make more sense there in between "*Execution KVP-encoded*" and "*Execution Steps*".
For consistency, the title could be "*Execution of Ad-hoc Workflow*"

docs/source/package.rst
```diff
@@ -681,6 +681,123 @@ Example multipart request structure:
...
+
+`Weaver` supports *ad-hoc* workflow execution, allowing users to deploy and execute a :term:`CWL` workflow in a single
```
fmigneault: Please add this:


```rst
# in references.rst (near corresponding definitions)

.. |ogc-api-proc-part3-cwl-ad-hoc-exec| replace:: :term:`CWL` *ad-hoc* workflow execution
.. _ogc-api-proc-part3-cwl-ad-hoc-exec: https://docs.ogc.org/DRAFTS/21-009.html#section_cwl_workflows
```

Then, reword as 
```rst
`Weaver` supports |ogc-api-proc-part3-cwl-ad-hoc-exec|_ from the |ogc-api-proc-part3|_ standard, which allows ...
```

docs/source/package.rst
```diff
@@ -681,6 +681,123 @@ Example multipart request structure:
...
+request to the ``POST /jobs`` endpoint using ``multipart/mixed`` or ``multipart/related`` content types. This
+eliminates the need for separate :ref:`deployment <proc_op_deploy>` and :ref:`execution <proc_op_execute>` steps
+when running one-time workflows or testing workflow definitions.
+
+The multipart request must contain:
+
+1. One or more :term:`CWL` parts (workflow and any dependent tools) with
+   ``Content-Profile: http://www.opengis.net/def/ogcapi-processes/2.0/process-description``
```
fmigneault: The `Content-Profile: http://www.opengis.net/def/ogcapi-processes/2.0/process-description` header would apply to cases where additional process metadata is provided (i.e.: https://github.com/crim-ca/weaver/issues/990)

Here, the CWL parts just need to have the relevant `application/cwl[+json|yaml]`.

Therefore, there could optionally be other "Process Description" parts using the indicated `Content-Profile`.

Consider adding a reference to `proc_ogc_api_multi_cwl` section by indicating that it follows the same concepts and procedure defined there.

docs/source/package.rst
```diff
@@ -681,6 +681,123 @@ Example multipart request structure:
...
+2. One execution request part with ``Content-Profile: http://www.opengis.net/def/ogcapi-processes/2.0/execute``
```
fmigneault: Make sure to update the URIs with HTTPS.

docs/source/package.rst
```diff
@@ -681,6 +681,123 @@ Example multipart request structure:
...
+.. code-block:: http
+
+    POST /jobs HTTP/1.1
+    Host: weaver.example.com
+    Content-Type: multipart/mixed; boundary="adhoc-boundary"
+
+    --adhoc-boundary
+    Content-Type: application/cwl+json
+    Content-Profile: http://www.opengis.net/def/ogcapi-processes/2.0/process-description
+
```
fmigneault: To be removed since that is not the corresponding profile for the CWL part contents.

If the code already supports dealing with multiple `application/json` of different `Content-Profile`, you can move this under a separte section like:

```http
--adhoc-boundary
Content-Type: application/json
Content-Profile: http://www.opengis.net/def/ogcapi-processes/2.0/process-description

{
  "id": "adhoc-workflow",
  "description": "Example ad-hoc workflow deployed and executed inline."
}
```

docs/source/package.rst
```diff
@@ -681,6 +681,123 @@ Example multipart request structure:
...
+        "DockerRequirement": {
+          "dockerPull": "alpine:latest"
+        }
+      },
+      "stdout": "output.txt"
+    }
+
+    --adhoc-boundary
+    Content-Type: application/cwl+json
+    Content-Profile: http://www.opengis.net/def/ogcapi-processes/2.0/process-description
```
fmigneault: idem

tests/processes/test_utils.py
```diff
@@ -1620,6 +1625,54 @@ def test_extract_multipart_start_parameter(self, content_type, expected):
         result = _extract_multipart_start_parameter(content_type)
         assert result == expected
 
+    @pytest.mark.parametrize("headers,expected", [
```
fmigneault: use `["headers", "expected"]` form

adjust the indents so it looks like:
```python
@pytest.mark.parametrize(
    ["headers", "expected"],
    [
        (...),
    ],
    id=[
        ...
    ],
)
```

tests/processes/test_utils.py
```diff
@@ -1620,6 +1625,54 @@ def test_extract_multipart_start_parameter(self, content_type, expected):
...
+        "content_type_profile_quoted",
+        "content_type_profile_angle_brackets",
+        "content_type_profile_with_charset",
+        "content_profile_precedence",
+        "no_profile",
+        "no_headers"
+    ])
+    def test_extract_multipart_profile(self, headers, expected):
+        """
+        Test _extract_multipart_profile handles various profile specification methods.
```
fmigneault: use ``` :func:`_extract_multipart_profile` ```

tests/processes/test_utils.py
```diff
@@ -1620,6 +1625,54 @@ def test_extract_multipart_start_parameter(self, content_type, expected):
...
+
+        Tests both RFC 8521 Content-Profile header and profile parameter in Content-Type header.
```
fmigneault: use ``` :rfc:`8521` ```, `Content-Profile` and `Content-Type`

tests/processes/test_utils.py
```diff
@@ -1808,3 +1859,461 @@ def test_create_multipart_deploy_multiple_workflows(self):
 
         # First workflow should be the main one
         assert "start=workflow-1" in content_type
+
+
+@pytest.mark.multipart
```
fmigneault, line 1868: Also add `@pytest.mark.oap_part3` and `@pytest.mark.job`.

tests/processes/test_utils.py
```diff
@@ -1808,3 +1859,461 @@ def test_create_multipart_deploy_multiple_workflows(self):
...
+
+    These tests cover the ad-hoc workflow submission feature where users can submit
+    both CWL packages and execution request in a single multipart request.
+    """
+
+    @pytest.mark.parametrize("error_case,expected_error", [
+        ("missing_execution", "Missing execution request"),
+        ("missing_cwl", "Missing CWL packages"),
+        ("multiple_execution", "Multiple execution requests"),
+        ("empty_content", "Content is not multipart format"),
```
fmigneault: According to codecov (https://github.com/crim-ca/weaver/pull/997#issuecomment-4984451216), one of the case partially convered is the empty check:
```python
+     if not interpreted_parts:
+         raise HTTPBadRequest(json={
```

Maybe another test case is missing here?

tests/processes/test_utils.py
```diff
@@ -1808,3 +1859,461 @@ def test_create_multipart_deploy_multiple_workflows(self):
...
+    ], ids=["missing_execution_request", "missing_cwl_packages", "multiple_execution_requests", "empty_content"])
+    def test_parse_multipart_job_execution_errors(self, error_case, expected_error):
+        """
+        Test various error conditions in multipart job execution parsing.
+        """
+        boundary = "----Boundary123"
+
+        if error_case == "missing_execution":
```
fmigneault: Because each `error_case` is a separate `if`, it seems easier to setup IMO if those were actually separate test cases instead of parameterized. There are only a few lines shared by them, so it is fine to duplicate in this case.

tests/processes/test_utils.py
```diff
@@ -1808,3 +1859,461 @@ def test_create_multipart_deploy_multiple_workflows(self):
...
+        try:
+            error_message = exc_info.value.json.get("title", "")
+            assert expected_error in error_message
+        except (AttributeError, ValueError):
+            assert expected_error in str(exc_info.value)
+
+    @pytest.mark.parametrize("use_profiles", [True, False], ids=["with_profiles", "fallback_detection"])
+    def test_parse_multipart_job_execution_basic(self, use_profiles):
+        """
+        Test parsing multipart job execution with/without explicit Content-Profile headers.
```
fmigneault: use code-quotes for header in docstring

tests/processes/test_utils.py
```diff
@@ -1808,3 +1859,461 @@ def test_create_multipart_deploy_multiple_workflows(self):
...
+
+        # Execution request
+        execution_request = json.dumps({
+            "inputs": {"message": "test"} if use_profiles else {"param": "value"},
+            "outputs": {"output": {"transmissionMode": "reference"}} if use_profiles else {},
+            "mode": "async"
+        } if use_profiles else {
+            "inputs": {"param": "value"},
+            "outputs": {}
+        })
```
fmigneault: Can more details be provided in the comment here?

I'm not sure to understand what triggers the fallback detection, or what the `param: value` is intended to represent?

The multiple mix of `use_profile` makes it hard to interpret as well. Won't the first 2 always be `true` since they need to be `true` to enter the "truthy" part of the later `if use_profiles`?

tests/processes/test_utils.py
fmigneault, line 1: `Content-Profile` for process description to be adjusted as per other comments.

There should be a test as well making sure that, if `Content-Profile: {sd.OGC_API_PROC_PROFILE_PROC_DESC_URI}` was indeed provided in a part along CWL contents, that it raises from schema valiation. When that header is provided, the contents should be checked against `sd.ProcessDeployment()` which does not include CWL definitions.

tests/processes/test_utils.py
```diff
@@ -1808,3 +1859,461 @@ def test_create_multipart_deploy_multiple_workflows(self):
...
+            "cwlVersion": "v1.2",
+            "class": "CommandLineTool",
+            "id": "test-tool",
+            "inputs": {},
+            "outputs": {}
+        }
+
+        exec_data = {"inputs": {"param": "value"}, "outputs": {}}
+
+        if error_case == "missing_execution":
```
fmigneault: rather than `error_case` as the parametrized input, it seems easier to have the `interpreted_parts` directly, and remove any `if/elif` in the test case

tests/processes/test_utils.py
```diff
@@ -1808,3 +1859,461 @@ def test_create_multipart_deploy_multiple_workflows(self):
...
+        exec_req, cwl_packages = _organize_job_execution_parts(interpreted_parts, None)
+
+        assert exec_req == exec_data
+        # $graph should be expanded into 2 packages
+        assert len(cwl_packages) == 2
+        cwl_ids = [pkg["id"] for pkg in cwl_packages]
+        assert "tool-1" in cwl_ids
+        assert "workflow-1" in cwl_ids
+
+    @pytest.mark.parametrize("error_case,expected_title", [
```
fmigneault: reformat as per https://github.com/crim-ca/weaver/pull/997/changes#r3604125606

tests/processes/test_utils.py
```diff
@@ -1808,3 +1859,461 @@ def test_create_multipart_deploy_multiple_workflows(self):
...
+            ]
+
+        with pytest.raises(HTTPBadRequest) as exc_info:
+            _organize_job_execution_parts(interpreted_parts, None)
+
+        assert expected_title in exc_info.value.json.get("title", "")
+
+    def test_organize_job_execution_parts_with_root_workflow_cid(self):
+        """
+        Test _organize_job_execution_parts reorders based on root workflow CID.
```
fmigneault: use `:func:`

tests/wps_restapi/test_jobs.py
```diff
@@ -3751,6 +3752,474 @@ def test_job_results_with_digest_multibase(self):
         expected_digest = compute_file_digest_multibase(test_file)
         assert output_result["digestMultibase"] == expected_digest
 
+    @pytest.mark.oap_part3
```
fmigneault, line 3758: add `multipart`, `job` and `functional` markers as well

same of other new tests below

tests/wps_restapi/test_jobs.py
```diff
@@ -3751,6 +3752,474 @@ def test_job_results_with_digest_multibase(self):
...
+    def test_job_ad_hoc_workflow_multipart_execution(self):
+        """
+        Test ad-hoc CWL workflow submission via multipart content to POST /jobs.
```
fmigneault: use code quotes for `POST /jobs`

tests/wps_restapi/test_jobs.py
```diff
@@ -3751,6 +3752,474 @@ def test_job_results_with_digest_multibase(self):
...
+
+        This validates that users can submit both the workflow definition (CWL) and
+        execution parameters (inputs/outputs) in a single multipart request to create
+        and execute a job without pre-deploying the process.
+
+        The multipart request contains:
+        - CWL workflow definition with Content-Profile: ogc-process-description
+        - Execution request with Content-Profile: ogc-execute-request
```
fmigneault: Headers with code quotes. Adjust profile for process descrition as per other comments.

tests/wps_restapi/test_jobs.py
```diff
@@ -3751,6 +3752,474 @@ def test_job_results_with_digest_multibase(self):
...
+            resp = mocked_sub_requests(
+                self.app, "post", "/jobs",
+                data=multipart_body,
+                headers={"Content-Type": content_type_header, "Accept": ContentType.APP_JSON},
+                only_local=True
+            )
+
+            # Verify the response
+            assert resp.status_code in [200, 201], (
+                f"Expected 200 or 201, got {resp.status_code}. "
```
fmigneault: refine 201
that is the expected result since async
same for others below

tests/wps_restapi/test_jobs.py
```diff
@@ -3751,6 +3752,474 @@ def test_job_results_with_digest_multibase(self):
...
+        assert sd.OGC_API_PROC_PROFILE_EXECUTE_URI in result["description"]
+
+    @pytest.mark.oap_part3
+    def test_job_ad_hoc_workflow_multipart_fallback_detection(self):
+        """
+        Test ad-hoc workflow submission with profile fallback detection.
+
+        When Content-Profile headers are omitted, the parser should detect:
+        - CWL content by 'class' field (Workflow, CommandLineTool)
+        - Execution request by 'inputs'/'outputs' fields
```
fmigneault: use code quotes

tests/wps_restapi/test_jobs.py
```diff
@@ -3751,6 +3752,474 @@ def test_job_results_with_digest_multibase(self):
...
+            assert process is not None, "Workflow should be deployed as a process"
+            assert process.identifier == process_id
+
+    @pytest.mark.oap_part3
+    def test_job_ad_hoc_workflow_multipart_unexpected_deployment_response(self):
+        """
+        Test ad-hoc CWL workflow submission when deployment returns an unexpected response type.
+
+        This validates that the system properly handles cases where the deployment operation
+        returns a response that is not HTTPCreated or HTTPOk.
```
fmigneault: use ``` :class:`HTTPCreated` or :class:`HTTPOk` ```

tests/wps_restapi/test_jobs.py
```diff
@@ -3751,6 +3752,474 @@ def test_job_results_with_digest_multibase(self):
...
+                    expect_errors=True
+                )
+
+                # Verify that we get a 400 Bad Request with the expected error message
+                assert resp.status_code == 400, f"Expected 400, got {resp.status_code}"
+                result = resp.json
+                assert "title" in result
+                assert result["title"] == "Unexpected deployment response"
+                assert "description" in result
+                assert "Ad-hoc workflow deployment did not return expected response" in result["description"]
```
fmigneault, line 4157: Does that response contain an `error` or similar detail to help users interpret what went wrong?

If no, it could be good to forward the received deployment error into the produced ad-hoc execution error response so they can debug their invalid process deploy definition.

If yes, that should be checked here to ensure it is provided.

tests/wps_restapi/test_jobs.py
```diff
@@ -3751,6 +3752,474 @@ def test_job_results_with_digest_multibase(self):
...
+                    expect_errors=True
+                )
+
+                # Verify that we get a 400 Bad Request with the expected error message
+                assert resp.status_code == 400, f"Expected 400, got {resp.status_code}"
+                result = resp.json
+                assert "title" in result
+                assert result["title"] == "Missing process ID"
+                assert "description" in result
+                assert "Ad-hoc workflow deployment did not return a process ID" in result["description"]
```
fmigneault: "provide" instead of "return" ?

docs/source/package.rst
```diff
@@ -681,6 +681,123 @@ Example multipart request structure:
...
+When an ad-hoc execution request is received:
+
+- The workflow (and any dependent tools) are automatically deployed with ``ad-hoc`` tagging
+- The job is immediately submitted for execution using the provided inputs
+- The workflow remains deployed after execution and can be reused or cleaned up later
+
+Example Ad-hoc Execution Request
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. code-block:: http
```
fmigneault: Consider moving this into https://github.com/crim-ca/weaver/tree/master/weaver/wps_restapi/examples and use a `literalinclude`. That will be useful for reuse in swagger definitions.

weaver/wps_restapi/swagger_definitions.py
```diff
@@ -7755,6 +7755,22 @@ class PostProcessJobsEndpointXML(PostJobsEndpointXML, LocalProcessPath):
...
+        missing=drop,
+        default=ContentType.MULTIPART_MIXED,
+        example=ContentType.MULTIPART_MIXED
+    )
+
+
+class PostJobsEndpointMultipart(ExtendedMappingSchema):
+    header = ExecuteHeadersMultipart()
+    querystring = LocalProcessQuery()
+    # No body schema - multipart content is parsed manually in the handler
```
fmigneault: Please add a concrete `ExecuteMultipart` example (see also https://github.com/crim-ca/weaver/pull/997/changes#r3604487411) for this class (request body). See the `PostJobsEndpointXML` for comparison.

We might need a partial definition of the content used in https://github.com/crim-ca/weaver/pull/997/changes#r3604487411 because that should contain just the "multipart body" portion (not including the top-most HTTP mehod/url/headers). 

For the `ExecuteMultipartBody`, I think it will work if it is a simple definition that derives from `ExtendedMappingSchema` with nothing in it. There shouldn't be any actual validation performed against it, it is just for rendering in SwaggerUI.

CHANGES.rst
```diff
@@ -12,6 +12,13 @@ Changes
 
 Changes:
 --------
+- Add support for ad-hoc `CWL` workflow execution through the ``POST /jobs`` endpoint using ``multipart/mixed``
+  or ``multipart/related`` content types. Users can now deploy and execute a `Process` in a single request by
+  submitting both the `CWL` workflow definition (with ``Content-Profile: http://www.opengis.net/def/ogcapi-processes/2.0/process-description``)
```
fmigneault: Needs adjustment as per other comments.

weaver/wps_restapi/jobs/jobs.py
```diff
@@ -216,6 +218,14 @@ def _job_list(_jobs):  # type: (Iterable[Job]) -> List[JSON]
     return Box(body, service=service, process=process)  # pass queries for contextual HTML elements
 
 
+@sd.jobs_service.post(
+    tags=[sd.TAG_EXECUTE, sd.TAG_JOBS, sd.TAG_PROCESSES],
+    content_type=list(ContentType.ANY_MULTIPART),
```
fmigneault: Technically, this will allow `multipart/form-data` that is not supported.
Employ a list of the supported ones explicitly.

weaver/wps_restapi/swagger_definitions.py
```diff
@@ -7755,6 +7755,22 @@ class PostProcessJobsEndpointXML(PostJobsEndpointXML, LocalProcessPath):
...
 
 
+class ExecuteHeadersMultipart(ExecuteHeadersBase):
+    # Override content_type to allow multipart types without strict validation
+    # Multipart headers include parameters (e.g., boundary=...) that vary per request
+    content_type = ContentTypeHeader(
+        missing=drop,
+        default=ContentType.MULTIPART_MIXED,
+        example=ContentType.MULTIPART_MIXED
+    )
```
fmigneault: Should have `validator=OneOf([...])` of the supported multipart variants.

While it is true the `boundary` is needed in their case, I'm not sure if they render correctly in SwaggerUI without the list of allowed values? I don't think they are actually validated using this schema, so the missing/partially defined `boundary` might be "good enough" for documentation. See if something similar to the following renders nicely in swagger and doesn't break validation:

```python
ContentTypeHeader(
    ...,
    validator=validator=OneOf([
        f"{ctype}; boundary=\"...\"" for ctype in 
        [ContentType.MULTIPART_MIXED, ContentType.MULTIPART_RELATED]
    ]
)
```

weaver/wps_restapi/jobs/jobs.py
```diff
@@ -243,8 +253,53 @@ def create_job(request):
...
+
+            deploy_body = deploy_response.json if hasattr(deploy_response, 'json') else deploy_response
+            proc_id = deploy_body.get("processSummary", {}).get("id")
+            if not proc_id:
+                raise HTTPBadRequest(json={
+                    "title": "Missing process ID",
+                    "description":
+                        f"Ad-hoc workflow deployment did not return a process ID. "
+                        f"Response: {deploy_body}",
+                })
```
fmigneault: A dedicated function in `jobs.utils` seems adequate. It only needs to return the `proc_id, execution_request`, which would be applied onto the below `request` overrides.

Consider renaming `execution_request` to `execution_body` since it doesn't contain all the request components (headers, queries, body), so it is misleading when reading the code.

weaver/wps_restapi/jobs/jobs.py
```diff
@@ -243,8 +253,53 @@ def create_job(request):
...
+
+            # Update request to use execution parameters for normal job submission flow
+            request._json = execution_request
+            request.json_body = execution_request
+            # Update content type so validate_job_json doesn't reject it
+            request.content_type = ContentType.APP_JSON
+            # Continue to normal job submission below with ad-hoc tag
```
fmigneault: no need for this comment, it is self-evident from code flow

weaver/processes/utils.py
```diff
@@ -760,19 +760,44 @@ def _extract_multipart_start_parameter(content_type):
     return None
 
 
+def _extract_multipart_profile(part):
+    # type: (Any) -> Optional[str]
```
fmigneault: Since `part.get()` is done first, looks like a `dict` needed instead of `Any`? Or is it an object from `email` module?

weaver/processes/utils.py
```diff
@@ -760,19 +760,44 @@ def _extract_multipart_start_parameter(content_type):
...
+    profile = part.get('Content-Profile', '').strip().strip('<>')
+    if profile:
+        return profile
+
+    # Try profile parameter in Content-Type header
+    content_type_full = part.get('Content-Type', '')
+    if 'profile=' in content_type_full:
+        profile_part = content_type_full.split('profile=')[1].split(';')[0].strip().strip('"').strip('<>')
+        if profile_part:
+            return profile_part
```
fmigneault: Could `get_response_profile` be reused? Maybe not as is, but with slight input parameter tweaks?

I'm expecting in the long run there will be many profile URIs, complexities and sources when parsing them (#993, #984), so I would rather have "profile parsing" logic more centralized in one place.

weaver/processes/utils.py
```diff
@@ -830,6 +855,105 @@ def _organize_deploy_parts(interpreted_parts, root_workflow_cid):
     return cwl_packages, process_description
 
 
+def _organize_job_execution_parts(interpreted_parts, root_workflow_cid):
+    # type: (List[Tuple[str, str, str, str, JSON]], Optional[str]) -> Tuple[JSON, List[CWL]]
+    """
+    Organize interpreted multipart parts into execution request and CWL packages for ad-hoc job execution.
+
+    :param interpreted_parts: ``list`` of interpreted parts (content_type, content_id, content_location, profile, data)
+    :param root_workflow_cid: Content-ID of root workflow from ``start`` parameter (if any)
```
fmigneault: code quotes for `Content-ID` and the tuple

weaver/processes/utils.py
```diff
@@ -830,6 +855,105 @@ def _organize_deploy_parts(interpreted_parts, root_workflow_cid):
...
+    :raises HTTPBadRequest: If execution request or CWL packages are missing or invalid
+    """
+    execution_request = None
+    deployment_parts = []
+    fallback_json_parts = []
+
+    for content_type_part, content_id, content_location, profile, part_data in interpreted_parts:
+        # Check if this is an execution request based on profile
+        if profile == sd.OGC_API_PROC_PROFILE_EXECUTE_URI:
+            if execution_request is not None:
```
fmigneault: Planing ahead, the `profile` should be a list where matching must be "any of" those URI.

weaver/processes/utils.py
```diff
@@ -830,6 +855,105 @@ def _organize_deploy_parts(interpreted_parts, root_workflow_cid):
...
+                })
+            execution_request = part_data
+        # Check if this is a deployment part based on profile or content
+        elif profile == sd.OGC_API_PROC_PROFILE_PROC_DESC_URI or (
+            isinstance(part_data, dict) and part_data.get("class") in ["CommandLineTool", "Workflow", "ExpressionTool"]
+        ) or (
+            isinstance(part_data, dict) and "cwlVersion" in part_data and "$graph" in part_data
+        ):
+            deployment_parts.append((content_type_part, content_id, content_location, profile, part_data))
+        # JSON/YAML parts without profile could be execution request (fallback)
```
fmigneault, line 889: for clairity, indicate that the YAML is expecte to have been pre-converted to JSON

weaver/processes/utils.py
```diff
@@ -830,6 +855,105 @@ def _organize_deploy_parts(interpreted_parts, root_workflow_cid):
...
+                "or include a JSON part with 'inputs' and/or 'outputs' fields."
+            ),
+        })
+
+    if not deployment_parts:
+        raise HTTPBadRequest(json={
+            "title": "Missing CWL packages",
+            "description": (
+                "Multipart request must contain at least one CWL package part for ad-hoc workflow execution. "
+                f"Use Content-Profile: {sd.OGC_API_PROC_PROFILE_PROC_DESC_URI} header "
```
fmigneault: to adjust as per other comments
hint toward the CWL media-types instead

weaver/processes/utils.py
```diff
@@ -830,6 +855,105 @@ def _organize_deploy_parts(interpreted_parts, root_workflow_cid):
...
+                "on CWL parts or ensure CWL content has 'class' field."
+            ),
+        })
+
+    # Extract CWL packages from deployment parts
+    cwl_packages = []
+    parts_by_cid = {}
+    for _, content_id, _, _, part_data in deployment_parts:
+        if isinstance(part_data, dict):
```
fmigneault: Should there be an extra check of `application/cwl[+json|+yaml]` here?

fmigneault (changes requested):
Great PR! Thanks for the updates. Some semantic adjustment needed for the process description profile URI that is misused.

weaver/wps_restapi/swagger_definitions.py
fmigneault: Not sure if it will work, but worth trying...

According to OpenAPI, multipart should be handled by user "part names" as the properties, and sub-schemas as the part's content (https://swagger.io/docs/specification/v3_0/describing-request-body/multipart-requests/).

So maybe using the following could work.

```python
class DeployCWLPart(AnyOfKeywordSchema):
    _any_of = [
        DeployCWL(),
        DeployCWLGraph(),
    ]


class ExecuteBodyMultipart(AnyOfKeywordSchema):
    # multipart are defined with properties in OpenAPI
    # see https://swagger.io/docs/specification/v3_0/describing-request-body/multipart-requests/
    deploy_cwl = DeployCWLPart(description="CWL tools and workflow parts.")
    process_meta = ProcessDeployment(drop=missing, description="Process description parts.")
    execute_body = Execute(description="Execution body.")


class PostJobsEndpointMultipart(ExtendedMappingSchema):
    header = ExecuteHeadersMultipart()
    querystring = LocalProcessQuery()
    body = ExecuteBodyMultipart()
```

tests/processes/test_utils.py
Nazim-crim: not sure how the content profile for process description should be adjusted, it's used in the code

weaver/processes/utils.py
Nazim-crim: object from email module typing adjusted

weaver/wps_restapi/swagger_definitions.py
Nazim-crim: Not able to put the boundary in the validator , used the copilot suggestion for stripping

weaver/processes/utils.py
Nazim-crim: String or list ?

weaver/processes/utils.py
Nazim-crim: parts with incorrect Content-Types don't get parsed and never reach that code.

tests/wps_restapi/test_jobs.py
Nazim-crim: adjust profile for process description in what way ?

tests/processes/test_utils.py
```diff
@@ -1630,8 +1683,6 @@ def test_interpret_multipart_part(self, content, content_type, should_parse):
         """
         Test _interpret_multipart_part with various content types and validity.
```
fmigneault: use ``` func:`_interpret_multipart_part` ```

tests/processes/test_utils.py
fmigneault: With latest updates, I think it is only missing the case validating a mismatching profile. In other words, if a known profile like `sd.OGC_API_PROC_PROFILE_PROC_DESC_URI` is provided for the wrong structure (eg: missing required parameters of the process description), that it raises from schema validation and fails the entire operation.

fmigneault (approved):

PR #998, opened 2026-07-17
improve OpenAPI examples for per-content-type process deployment

fmigneault:
## Changes

API rendering used to return the examples in mixed order/combination per request body content-type because they were are combined under one schema.

Now, they properly offer all media-types (some of which were missing), and picking one will offer corresponding OGC/CWL representations.

For now, JSON/YAML are still returned together. Otherwise we would need to create even more schema class combinations to actually represent each one separately. We keep it relatively simple with that simplification.

{AACEE60A-E1FC-4015-B920-AEC0F21842D8}

{D373D01D-D4E0-4BA6-B3C8-36A29375D7AD}

{4CC53AA4-1057-4F62-B5F7-3B9039279011}

{CB9918FD-5941-4868-AA4B-D5F27CE20DAD}

### DIFF
--- a/weaver/wps_restapi/swagger_definitions.py
+++ b/weaver/wps_restapi/swagger_definitions.py
@@ -7558,7 +7558,7 @@ class Deploy(OneOfKeywordSchema):
     ]
 
 
-class DeployContentType(ContentTypeHeader):
+class DeployContentTypeAny(ContentTypeHeader):
     example = ContentType.APP_JSON
     default = ContentType.APP_JSON
     validator = OneOf([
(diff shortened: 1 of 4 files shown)


weaver/wps_restapi/swagger_definitions.py
```diff
@@ -7573,19 +7573,52 @@ class DeployContentType(ContentTypeHeader):
...
+class DeployContentTypeCWL(ContentTypeHeader):
+    example = ContentType.APP_JSON
+    default = ContentType.APP_JSON
+    validator = OneOf([
+        ContentType.APP_CWL,
+        ContentType.APP_CWL_JSON,
+        ContentType.APP_CWL_YAML,
+        ContentType.APP_CWL_X,
+        ContentType.APP_YAML,
+        ContentType.APP_JSON,
```
Nazim-crim, line 7596: Shouldnt this be already covered with
```
ContentType.APP_CWL_JSON,
ContentType.APP_CWL_YAML,
```
Why do we need those two

weaver/wps_restapi/swagger_definitions.py
```diff
@@ -7594,16 +7627,32 @@ class PostProcessesEndpoint(ExtendedMappingSchema):
...
+class PostProcessesEndpointOGC(PostProcessesEndpoint):
+    header = DeployHeadersOGC(description="Headers employed for process deployment.")
+    querystring = FormatQuery()
+    body = DeployBodyOGC()
+
+
+class PostProcessesEndpointCWL(PostProcessesEndpoint):
+    header = DeployHeadersCWL(description="Headers employed for process deployment.")
+    querystring = FormatQuery()
+    body = DeployBodyCWL()
```
Nazim-crim: The description  could reflect if its CWL or OGC headers for more context

Nazim-crim (dismissed):

weaver/wps_restapi/swagger_definitions.py
fmigneault: They are different headers (`application/json` vs `application/cwl+json`).
If both are not included, a request using the missing once would not validate (406 response) since `sd.DeployContentTypeCWL.validator.choices` is used in the decorator of the endpoint to validate `content_type`.

PR #968, opened 2026-05-22
Add support for multi deployment using workflow cwl

Nazim-crim:
- fixes https://github.com/crim-ca/weaver/issues/717
- fixes https://github.com/crim-ca/weaver/issues/874
- relates to https://github.com/opengeospatial/ogcapi-processes/pull/588

-support multiple executionUnits CWL entries
This is the current method already supported, but it stops at the first iteration. It should allow multiple ones, but only if a CWL-like media-type is provided, such that single-deployment of other types (e.g.: application/json or application/ogcapppkg) do not lead to ambiguities.
- Add support for multipart `Process` deployment using ``multipart/related`` or ``multipart/mixed`` content types,
  allowing simultaneous upload of multiple `CWL` files (e.g.: ``class: Workflow`` and ``class: CommandLineTool``)
  along with optional `Process` description metadata 
  - relates to #56
  - resolves #717
  - resolves #874
- Add support for `CWL` ``$graph`` representation with multiple entries, enabling deployment of workflows with
  embedded step definitions without manual preprocessing
  (resolves #56).
- Add validation during multipart deployment to ensure at least one ``class: Workflow`` is present and properly
  identified as the main workflow from media-type hints or ``Content-ID`` references.
- Add `CLI` support for deploying multiple `CWL` files by repeating ``--cwl`` arguments to automatically generate
  ``multipart/related`` request payloads, simplifying deployment of workflows with multiple step definitions.
  The `CLI` accepts a mixture of local `CWL` file paths (as `JSON` or `YAML`) and remote URL references.
  The server handles ``Content-Type`` media-type detection and ``Content-Location`` header processing for external
  `CWL` files to fetch during deployment.

Issue #717, opened 2024-09-18
Support Multipart CWL Deployment


fmigneault:
## Description

When a CWL consists of a `class: Workflow` composed of multiple CWL files, the deployment process can be painful, since users must deploy each child step (to ensure they resolve without missing dependencies) before deploying the final root workflow. 

Furthermore, although CWL can be deployed on their own, embedding it within the `executionUnit` allows the deployment to provide additional/complementary metadata and process description details. However, embedding the CWL within the process description must be done manually by the user, or using the CLI that happens to offer this capability.

Instead, `Content-Type: multipart/*` could be used. This would allow uploading all workflows steps simultaneously, as separate CWL parts, as well as any additional process description metadata. Using `multipart/related` and/or `multipart/mixed` (either in conjunction or separately), it is also possible to employ `Content-ID` (https://datatracker.ietf.org/doc/html/rfc2392.html) to provide relationships between the various child-CWL and their respective descriptions. It is also possible to use a nested hierarchy, as follows: 

- ``multipart/mixed``
  - root CWL workflow `application/cwl+json`
  - ``multipart/related``
    - step1 CWL CommandLineTool `application/cwl+json`
    - step1 ProcessDescription `application/json`
  - step2 CWL CommandLineTool `application/cwl+json`

## To Do

- [ ] Support the ``multipart/*`` process deployment 
  - [ ] Support for a plain CommandLineTool along its ProcessDescription metadata
  - [ ] Support for a workflow, which must deploy each step first, and then the root workflow
- [ ] Supported nested ``multipart/*``
- [ ] Consider `Content-ID` to match corresponding CWL and ProcessDescription definitions (since they are not necessarily next to each other in the contents)
- [ ] Consider `{cid:[...]}` cross-references if needed
- [ ] Consider if `multipart/related` must be enforced as top level ?  https://datatracker.ietf.org/doc/html/rfc2387
  - The definition `Content-Type: multipart/related; type="..."; start="<cid>` is available in the top-level of the request, which can indicate which is the root workflow.
- [ ] Consider `Content-Disposition` of each part for relevant CWL file references used for `run:` of the workflow
- [ ] Update the CLI to allow a list of CWL, which would generate the relevant multipart contents to deploy them  

## References

- n/a

PR #1002, opened 2026-08-18
fix deploy openapi for multipart

fmigneault:
## Detail

The decorator for `multipart/*` combinations was missing from https://github.com/crim-ca/weaver/pull/998
While adding this, it was identified that the `Content-Type` parameter was not containing all possible options, because it simply registered the values from the first view decorator.

This fixes the definition to obtain the valid combinations, and adds a missing example for multipart.

### Toggle Content body per Content-Type
image

### Example multipart 
image

### Aggregated `Content-Type` parameter with "*Try it out*" submission
image

## Changes

- Fix rendering of `OpenAPI` definitions for ``POST /processes`` deployment such that each indicated ``Content-Type``   correctly provides an example for ``multipart/*`` representation of a `Workflow`.
- Fix `OpenAPI` combination of ``Accept`` and ``Content-Type`` header values from all `OGC Application Package`, `CWL`   and ``multipart/*`` schemas when aggregated under the *request parameters* of the `Swagger-UI` representation.

## References

- Continuation of adjustments from #998 
- Relates to #717 
- Relates to #968

### DIFF
--- a/weaver/wps_restapi/swagger_definitions.py
+++ b/weaver/wps_restapi/swagger_definitions.py
@@ -7558,21 +7558,6 @@ class Deploy(OneOfKeywordSchema):
     ]
 
 
-class DeployContentTypeAny(ContentTypeHeader):
-    example = ContentType.APP_JSON
-    default = ContentType.APP_JSON
-    validator = OneOf([
-        ContentType.APP_JSON,
-        ContentType.APP_CWL,
-        ContentType.APP_CWL_JSON,
-        ContentType.APP_CWL_YAML,
-        ContentType.APP_CWL_X,
-        ContentType.APP_OGC_PKG_JSON,
-        ContentType.APP_OGC_PKG_YAML,
-        ContentType.APP_YAML,
-    ])
-
-
 class DeployContentTypeOGC(ContentTypeHeader):
     example = ContentType.APP_JSON
     default = ContentType.APP_JSON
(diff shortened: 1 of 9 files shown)


weaver/wps_restapi/swagger_definitions.py
```diff
@@ -7612,9 +7616,9 @@ class DeployHeadersCWL(RequestHeadersBody):
     content_type = DeployContentTypeCWL()
 
 
-class PostProcessesEndpoint(ExtendedMappingSchema):
-    header = DeployHeadersAny(description="Headers employed for process deployment.")
-    querystring = FormatQuery()
+class DeployHeadersMultipart(RequestHeadersBody):
+    x_auth_docker = XAuthDockerHeader()
```
Nazim-crim, line 7620: Why do we need to specificy this Header instead of ANY. I dont see it being used

Nazim-crim (changes requested):

weaver/wps_restapi/swagger_definitions.py
fmigneault: You mean the `x_auth_docker` or the `DeployHeadersMultipart` class?

Nazim-crim: Both i only see it used in comment :     # header = DeployHeadersMultipart()  # see 'PostProcessesEndpoint'

fmigneault: Indeed. They are not used anymore, but I left them so it makes sense with that reference and we don't re-add it by mistake.

